PR #3, opened 2018-05-09
Fixed submodule path for DroneCore-Proto

xvzf:


JonasVautherin:
Thanks for that!

Let's continue the discussion about the Python interface [here](https://github.com/dronecore/DroneCore/issues/388). I'd just like to make sure we agree before we start implementing more here :-).

PR #12, opened 2018-05-23
[WIP] Implementation of the actual python bindings for DroneCore

xvzf:
This is still a WIP!

Basic concept is using the bindings like this:

```Python
from dronecore import Backend, Action

# Connect to a dronecore backend running on the local machine
core = Core(host="127.0.0.1")

# Initialize the Action plugin
action = Action(core)

# E.g. arm the copter
action.arm().subscribe(on_completed=lambda: print("ARMED!"),
                       on_error=lambda _: print("Could not arm the system!"))
```


dronecore/backend.py
```diff
@@ -0,0 +1,90 @@
+""" Backend """
```
JonasVautherin: Wording question: isn't this what we call `Core`? What we call the "backend" is really just the gRPC server side (C++). So the `Core` spins up a backend.

xvzf: Okay, Let's rename that to Core! :-)

dronecore/plugins/action.py
```diff
@@ -65,64 +64,66 @@ def takeoff(self):
...
+                observable.on_error(response)
+
+        return self._create_call_observable(f)
+
+#   @TODO seems to not be working, action_result is not available
+    def get_takeoff_altitude(self):
+        """
+        Takeoff
+
+        :returns: Ovservable
```
JonasVautherin: Ovservable -> Observable :)

xvzf: Thanks!

xvzf:
> I think I like it :-). Can we merge that in this state already?

I'd actually like to get streaming working properly before - also I have some changes that are not commited yet! Let me check this afternoon, maybe then we can merge!

JonasVautherin:
No worries, it's not urgent :)

LorenzMeier:
It's always urgent! 8). I'd love to move the PX4 CI over to this!

dronecore/plugin_manager.py
```diff
@@ -60,24 +59,19 @@ def _spinup_backend(self):
         """
         Spinup a backend and connect to it
         """
+        #: backend is running on localhost
+        self.port = "127.0.0.1"
```
JonasVautherin: I'm not sure I understand the worgind here: default port is 50051, and address is localhost.

xvzf: That's indeed wrong. I haven't tested the function yet as I am not sure how we will ship the backend with the python library!

JonasVautherin: Right. Well, the simple assumption is "the backend is already running on an \<IP:port\> somewhere". And on top of that, we can provide an interface that runs a backend locally. But it may be that the backend runs on a different machine (e.g. directly on the drone), so at least one should be able to connect to \<IP:port\>.

In Swift, that's currently the only interface (e.g. `let core = Core("localhost", 50051)`).

xvzf: That is possible? Have a look at ``__init__`

```
import dronecore

# Connect to a backend
remote_backend = dronecore.connect(host="192.168.178.56", port=50051)

# Start a backend local and connect
local_backend = dronecore.connect()
```

JonasVautherin: yeah, I know :). I was just answering to your comment, but maybe I missed your point actually ;-).

xvzf:
@JonasVautherin FYI

Basic calls and Streams are working just fine, I kept the code pretty consistent so we can build the auto generator ontop of it!

@hamishwillee I don't think the interface will change - I haven't worked with the Swift bindings yet but tried to kept it as close as possible, can you compare and give your OK? :-)

dronecore/plugins/telemetry.py
```diff
@@ -14,40 +152,99 @@ def __init__(self, *args, **kwargs):
...
-                    pass
-            except Exception as e:
-                observable.on_error(e)
+        self._armed_observable = self._create_stream_observable(
+                response.is_armed for response in stream)
 
-            observable.on_completed()
+    def _create_position_observable(self):
+        """
+        Create the position ovservable
```
JonasVautherin: ovservable -> observable :). In all the other functions, too ;-)

dronecore/plugins/telemetry.py
```diff
@@ -1,7 +1,145 @@
+# -*- coding: utf-8 -*-
 from .._base import Base
 from ..generated import telemetry_pb2, telemetry_pb2_grpc
 
 
+class Position(object):
```
JonasVautherin: Nice!

dronecore/plugins/telemetry.py
xvzf: damnit! Thanks!

xvzf:
@JonasVautherin I propose we merge and take it as a baseline for the autogenerator - I can't think of anything that has to be changed.

PR #15, opened 2018-08-20
Moving to dronecode sdk

xvzf:
@julianoes should be it!


README.rst
```diff
@@ -11,13 +11,13 @@ Clone the repo (or your fork) and recursively update submodules:
...
 
-    git clone https://github.com/dronecore/DroneCore-Python.git --recursive
+    git clone https://github.com/dronecore/DronecodeSDK-Python.git --recursive
 
 This can be done as 2 steps, as shown below.
 
 .. code-block:: bash
 
-    git clone https://github.com/<your fork>/DroneCore-Python.git
+    git clone https://github.com/<your fork>/DronecodeSDK-Python.git
```
julianoes: I would probably remove the `<your fork>`.

hamishwillee: I like the `<your fork>` but it should be applied consistently - ie in the previous single-step line too.

julianoes: But we are not actually trying to promote forks...

hamishwillee: OK. I tend to assume that the model for contribution is that you branch on your fork and then submit a PR there. If not, then fair enough.

julianoes: Hm, I see, for development. You're right, and that's what this is for.

julianoes: I was thinking more about this and found that actually, we should not have to explain these things. Developers either know this or should google "how to make a github pull request".

hamishwillee: I don't feel strongly about that, but I do about the inconsistency. More important, the fact that the repo moved was bugging me. I have fixed it how @julianoes would like it in https://github.com/Dronecode/DronecodeSDK-Python/commit/d20e1be3b26594a81668df5e769a727cc1a1d293

julianoes: Thx.

PR #22, opened 2018-10-08
Update autogenerator

JonasVautherin:
One issue we have with the `NameParser` is that words like `GPSPosition` become `g_p_s_position`, so we should probably do something like taking the first and last capital letters when we have more than 1 (would be G and P here, for `gps_position`).

Secondly, @xvzf you mentioned I should use `connect` in my example, but I don't understand what you meant. Can you comment on this PR?


examples/takeoff_and_land.py
```diff
@@ -0,0 +1,23 @@
+import asyncio
+
+from dronecode_sdk import (AsyncPluginManager,
```
xvzf: I actually added a `connect` function in [__init__.py](https://github.com/Dronecode/DronecodeSDK-Python/blob/master/dronecode_sdk/__init__.py) which setups the default plugins and connects to the backend.

You can use it like this:
```Python
from dronecode_sdk import connect as dronecode_sdk_connect

copter = dronecode_sdk_connect(host="127.0.0.1")

# And inside your funcions you can access the plugins as a simple property, e.g. action:
await copter.action.arm()
```

examples/takeoff_and_land.py
```diff
@@ -0,0 +1,23 @@
...
+                           Action)
+
+manager = AsyncPluginManager(host="127.0.0.1")
+action = Action(manager)
+
+async def run():
+
+    arm_result = await action.arm()
+    print("Arm result: " + str(arm_result[1].action_result.result_str))
```
xvzf: No need to call `str` afaik

examples/takeoff_and_land.py
```diff
@@ -0,0 +1,23 @@
...
+    takeoff_result = await action.takeoff()
+    print("Takeoff result: " + takeoff_result[1].action_result.result_str)
```
xvzf: It's fine here, you see? 😛

other/templates/call.j2
```diff
@@ -1,10 +1,10 @@
 
-async def {{ name }}(self):
+async def {{ name.lower_camel_case }}(self):
     """ Generated by dcsdkgen
 
     :returns: Tuple[Success, Response]
     """
 
-    response = await self._stub.{{ name }}({{ plugin_name.lower() }}_pb2.{{ name }}Request())
+    response = await self._stub.{{ name.upper_camel_case }}({{ plugin_name.lower_snake_case }}_pb2.{{ name.upper_camel_case }}Request())
```
xvzf: This looks so much better! 👍

examples/takeoff_and_land.py
JonasVautherin: Right, it was `str(arm_result[1])` before, and I did not realize that. Thanks!

examples/takeoff_and_land.py
```diff
@@ -0,0 +1,21 @@
+import asyncio
+
+from dronecode_sdk import connect as dronecode_sdk_connect
+
+drone = dronecode_sdk_connect(host="127.0.0.1")
```
xvzf: 👍 Looks good!

PR #23, opened 2018-10-15
Add telemetry examples

JonasVautherin:
Fixed a few more bugs in the generation, which shows that it will be helpful to generate unit tests (but that will come).

### DIFF
--- a/other/templates/enum.j2
+++ b/other/templates/enum.j2
@@ -8,4 +8,4 @@ class {{ name.upper_camel_case }}:
             {%- for value in values %}
                 {{ loop.index - 1 }}: "{{ value.uppercase }}",
             {%- endfor %}
-            }.get(response.{{ name.lower_snake_case }}, default=None)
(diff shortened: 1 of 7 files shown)


other/templates/enum.j2
```diff
@@ -8,4 +8,4 @@ class {{ name.upper_camel_case }}:
             {%- for value in values %}
                 {{ loop.index - 1 }}: "{{ value.uppercase }}",
             {%- endfor %}
-            }.get(response.{{ name.lower_snake_case }}, default=None)
+            }.get(response.{{ name.lower_snake_case }}, None)
```
JonasVautherin: This resulted in a runtime error: `TypeError: get() takes no keyword arguments`

xvzf: Didn't get that! :+1:

julianoes:
Don't forget to delete the branches, please :smile:.

PR #24, opened 2018-10-18
Different approach to create the tasks for the eventloop

xvzf:
@JonasVautherin I adapted the example to give an example on how to setup coroutines in larger scale applications!

### DIFF
--- a/examples/telemetry_takeoff_and_land.py
+++ b/examples/telemetry_takeoff_and_land.py
@@ -6,6 +6,7 @@ drone = dronecode_sdk_connect(host="127.0.0.1")
 
 
 async def run():
+    """ Arms the system, initiates a takeof and finally lands the system """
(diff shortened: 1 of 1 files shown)


examples/telemetry_takeoff_and_land.py
```diff
@@ -6,6 +6,7 @@
 
 
 async def run():
+    """ Arms the system, initiates a takeof and finally lands the system """
     arm_result = await drone.action.arm()
     print("-- Arm result: " + arm_result[1].action_result.result_str)
```
xvzf: I will do something about the `arm_result[1].action_result.result_str`, I don't like to type this and it should be the default output!

xvzf: This PR can be merged, I'll go for a new one!

PR #26, opened 2018-10-18
add grpcio-tools to requirements

JonasVautherin:
`run_protoc.sh` fails without `grpcio-tools` installed.


xvzf:
If any one has the same issue:

Runtime dependencies are in `requirements.txt`
Development dependencies are in `requirements-dev.txt`
Testing dependencies are in `requirements-test.txt`

JonasVautherin:
Just for the record:

```pip install -r requirements-dev.txt -e .```

PR #27, opened 2018-10-18
[WIP] Sync API based on the asyncio implementation

xvzf:
Goal is to wrap the asyncio implementation in "normal" classes and provide an easy to use and understand sync API


JonasVautherin:
Let's close this as it is outdated. We can still get inspiration from it at some point if we make a sync API :slightly_smiling_face:.

PR #31, opened 2018-11-06
Dispose streams when closed on the client side

JonasVautherin:
Was previously leaking threads...

This will allow us to generate polling for the "simple api", with something like:

    def position(self):
        async def print_position_aio():
            position_gen = self.telemetry.position()
            position = await position_gen.__anext__()
            position_gen.aclose()

            return position

        return self.loop.run_until_complete(print_position_aio())


other/templates/stream.j2
```diff
@@ -1,10 +1,15 @@
 
 async def {{ name.lower_snake_case }}(self):
     """ Generated by dcsdkgen """
-    async for response in self._stub.Subscribe{{ name.upper_camel_case }}(
-                {{ plugin_name.lower_snake_case }}_pb2.Subscribe{{ name.upper_camel_case }}Request()):
+    {{ name.lower_snake_case }}_stream = self._stub.Subscribe{{ name.upper_camel_case }}(
+                {{ plugin_name.lower_snake_case }}_pb2.Subscribe{{ name.upper_camel_case }}Request())
```
julianoes: That's so cool :smile: :snake: :camel:

other/templates/stream.j2
```diff
@@ -1,10 +1,15 @@
...
+    async for response in {{ name.lower_snake_case }}_stream:
+        try:
         {%- if return_type.is_primitive %}
             yield response.{{ return_name.lower_snake_case }}
         {%- else %}
             yield {{ return_type.name }}.parse_response(response)
-        {%- endif -%}
+        {%- endif %}
+        finally:
+            {{ name.lower_snake_case }}_stream.cancel()
```
xvzf: This is great! :+1:

PR #35, opened 2018-11-13
Fix stream cancelling and improve example

JonasVautherin:
* `_response_success` was hard-coded for action
* streams where not cancelled correctly (finally was called directly after the first element)


examples/takeoff_and_land.py
```diff
@@ -9,15 +9,20 @@
 
 async def run():
 
-    arm_result = await drone.action.arm()
-    print("Arm result: " + arm_result[1].action_result.result_str)
-    takeoff_result = await drone.action.takeoff()
-    print("Takeoff result: " + takeoff_result[1].action_result.result_str)
+    arm_status, _ = await drone.action.arm()
```
julianoes: Shouldn't this be called `arm_result`?

examples/takeoff_and_land.py
```diff
@@ -9,15 +9,20 @@
...
+        return
+
+    takeoff_result, _ = await drone.action.takeoff()
+    print(f"-- Takeoff result: {takeoff_result}")
 
     await asyncio.sleep(5)
 
-    land_result = await drone.action.land()
-    print("Land result: " + land_result[1].action_result.result_str)
+    land_result, _ = await drone.action.land()
```
julianoes: What's the second return argument?

JonasVautherin: The error message

julianoes: As in a string instead of enum?

julianoes: Ah that makes sense.

PR #37, opened 2018-12-09
Add examples

JonasVautherin:
* Add mission example
* Improve `telemetry_takeoff_and_land`


examples/mission.py
```diff
@@ -0,0 +1,59 @@
...
+    my_items = []
+    my_items.append(MissionItem(47.398039859999997, 8.5455725400000002, 25, 10, True, 0, 0, MissionItem.CameraAction.NONE, 0))
+    my_items.append(MissionItem(47.398036222362471, 8.5450146439425509, 25, 10, True, 0, 0, MissionItem.CameraAction.NONE, 0))
+    my_items.append(MissionItem(47.397825620791885, 8.5450092830163271, 25, 10, True, 0, 0, MissionItem.CameraAction.NONE, 0))
+
+    mission_items = MissionItems(my_items)
+    upload_mission_status, _ = await drone.mission.uploadMission(mission_items)
+    print(f"-- Upload mission: {upload_mission_status}")
+
+    await drone.mission.setReturnToLaunchAfterMission(True)
```
julianoes: You need to set this one before you upload the mission.

examples/mission.py
```diff
@@ -0,0 +1,59 @@
...
+
+    arm_status, _ = await drone.action.arm()
+    print(f"-- Arm result: {arm_status}")
+
+    start_mission_status, _ = await drone.mission.startMission()
+    print(f"-- Start mission: {start_mission_status}")
+
+    await drone.mission.setCurrentMissionItemIndex(1)
```
julianoes: So you're skipping the first one on purpose?
That's not that intuitive.

examples/mission.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def setup_tasks():
+    asyncio.ensure_future(run())
+    asyncio.ensure_future(print_mission_progress())
+
+
+if __name__ == "__main__":
+    setup_tasks()
+    asyncio.get_event_loop().run_until_complete(observe_is_in_air())
```
julianoes: Nice!

examples/mission.py
julianoes: Yep :smile:

examples/mission.py
JonasVautherin: Right, I guess I was testing it, but it doesn't have to be in the example.

julianoes: Or it can be but then it's better with a comment about why you're doing it.

examples/mission.py
```diff
@@ -0,0 +1,57 @@
...
+
+from dronecode_sdk import connect as dronecode_sdk_connect
+from dronecode_sdk import (MissionItem, MissionItems)
+
+drone = dronecode_sdk_connect(host="127.0.0.1")
+
+
+async def run():
+    my_items = []
+    my_items.append(MissionItem(47.398039859999997, 8.5455725400000002, 25, 10, True, 0, 0, MissionItem.CameraAction.NONE, 0))
```
xvzf: Maybe split this up as as multi-line and add comments after each number

examples/mission.py
```diff
@@ -0,0 +1,57 @@
...
+        print(f"Mission progress: {mission_progress.current_item_index}/{mission_progress.mission_count}")
+
+
+async def observe_is_in_air():
+    """ Monitors whether the drone is flying or not and
+    returns after landing """
+
+    was_in_air = False
+
+    async for is_in_air in drone.telemetry.in_air():
```
xvzf: We could add some helper functions in the sdk, e.g. `exit_on_land` - what do you think about it?

examples/mission.py
julianoes: Best would be named arguments.

JonasVautherin: Agreed for named arguments, but that's not in the scope of this PR :+1:

examples/mission.py
JonasVautherin: I agree that this is a bit boilerplate, but I'm a bit nervous about adding stuff like that into the SDK :laughing:.

I'd say that for now, we leave it to the user.

examples/mission.py
JonasVautherin: > Maybe split this up as as multi-line and add comments after each number

Do we _have_ to :sweat_smile:? It looks scary to me when it takes 10 lines. At least here is reads like "a bunch of arguments that should be named arguments" :laughing:.

julianoes:
Please clean up the branch after merging, thanks.

PR #39, opened 2018-12-29
remove --user with pip3 install

Huibean:
remove --user in case ~/.local/ not in env PATH


xvzf:
Same here. we added the if condition intended so we can use both options of installing it.

xvzf:
I'll close this for now

PR #40, opened 2019-01-07
Add support for return values

JonasVautherin:
* Enable Calibration and Info plugins
* Add example for Info plugin (examples/firmware_version.py)
* Add support for return values (e.g. to actually get the version from the info plugin, instead of only querying it and never receiving it)
* Improve API by hiding gRPC (we now generate our own types, as in Swift)

### DIFF
--- a/examples/takeoff_and_land.py
+++ b/examples/takeoff_and_land.py
@@ -9,19 +9,15 @@ drone = dronecode_sdk_connect(host="127.0.0.1")
 
 async def run():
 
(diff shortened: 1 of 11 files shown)


examples/takeoff_and_land.py
```diff
@@ -9,19 +9,15 @@
 
 async def run():
 
-    arm_result, _ = await drone.action.arm()
+    arm_result = await drone.action.arm()
```
julianoes: Right, I think I like that better.

PR #41, opened 2019-01-07
Raise exception when result is not SUCCESS

JonasVautherin:
This is making things clearer IMO: with error codes, the risk is to get undefined behavior by not checking an error code.

It doesn't have any consequence on the other language bindings, as they already work with exceptions. So it is really a choice here, and I believe exceptions are more pythonic.

One thing, though, is that this breaks the Calibration plugin. Because only `SUCCESS` is currently considered as a success. Calibration is using `IN_PROGRESS` and `INSTRUCTION`, but those should disappear and become `NEXT`, to be consistent with our streaming approach everywhere. So in the future, whatever is not `SUCCESS` or `NEXT` will raise an exception.


xvzf:
LGTM! Should we merge already or wait for a fix regarding the `CALIBRATION` plugin

JonasVautherin:
I would merge already, I will work on the `NEXT` today.

PR #45, opened 2019-02-12
Update protos

JonasVautherin:
* Fix mission example
* Handle repeated fields in call.j2 and request.j2

### DIFF
--- a/examples/mission.py
+++ b/examples/mission.py
@@ -3,21 +3,47 @@
 import asyncio
 
 from dronecode_sdk import connect as dronecode_sdk_connect
-from dronecode_sdk import (MissionItem, MissionItems)
+from dronecode_sdk import (MissionItem)
 
 drone = dronecode_sdk_connect(host="127.0.0.1")
 
 
 async def run():
-    my_items = []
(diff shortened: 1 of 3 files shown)


examples/mission.py
```diff
@@ -3,21 +3,47 @@
...
-    my_items.append(MissionItem(47.398039859999997, 8.5455725400000002, 25, 10, True, 0, 0, MissionItem.CameraAction.NONE, 0))
-    my_items.append(MissionItem(47.398036222362471, 8.5450146439425509, 25, 10, True, 0, 0, MissionItem.CameraAction.NONE, 0))
-    my_items.append(MissionItem(47.397825620791885, 8.5450092830163271, 25, 10, True, 0, 0, MissionItem.CameraAction.NONE, 0))
+    mission_items = []
+    mission_items.append(MissionItem(47.398039859999997,
+                                     8.5455725400000002,
+                                     25,
+                                     10,
+                                     True,
+                                     float('nan'),
```
julianoes: You could also use: `math.nan`
https://docs.python.org/3/library/math.html#math.nan

JonasVautherin: But then I need to `import math` :sweat_smile:. Not sure what's best. I guess that's fine for this example.

julianoes: Up to you.

Issue #20, opened 2018-09-13
Packaging with backend


JonasVautherin:
Summarizing slack discussions with @xvzf @julianoes @mrpollo.

The python wrappers connect to the `backend` (a C++ binary), which in turns connects over mavlink to the drone. Using the python wrappers therefore requires to have an instance of the `backend` running. Building the `backend` is not trivial and not convenient on some environments (typically a companion computer without a build toolchain and with low computational power).

The question here is the following: **how do we package the python wrappers together with the `backend`, so that a user can simply `pip install` the SDK and start using it?**

Because the `backend` is a C++ library, it needs to be built for the different supported platforms (i.e. we cannot have one universal binary that would work everywhere).

We thought about a few solutions:

1. We give the responsibility to the user (who can build or install the backend *somehow*), and distribute the python wrappers over `pip`.
2. We build the `backend` for the platforms we support and distribute that through the platform's package manager (be it `apt`, `homebrew`, etc).
3. We build the `backend` for the platforms we support and distribute that with `pip`.

We tend to like the last two solutions better, because it makes it easier for a user to get started. 2) requires us to maintain different package managers, while 3) goes exclusively through `pip`.

One question comes with 3), though: can/should we make a binary that would work for all (/most) linux distributions? It would imply statically linking all the system libraries that are not common to all (/most) linux distributions.

xvzf:
Just a random thought - don't ship via pip at all, distribute debian packages for Ubuntu/Debian, rpms for Fedora/Redhat and an AUR Package for archlinux

julianoes:
> Just a random thought - don't ship via pip at all, distribute debian packages for Ubuntu/Debian, rpms for Fedora/Redhat and an AUR Package for archlinux

I'd prefer pip but I can see how that could be a possibility.

xvzf:
> I'd prefer pip but I can see how that could be a possibility.

That would be also my preferred way but that depends on statically linking the backend

hamishwillee:
> We give the responsibility to the user (who can build or install the backend somehow), and distribute the python wrappers over pip.

Definitely need to supply simple solution of pre-built binaries "somehow".

But even so we should provide clear instructions on how to set this up manually for unsupported platforms.

hamishwillee:
[...]

PR #58, opened 2019-04-03
Support embedding of the grpc server binary (backend_bin)

JonasVautherin:
I modified the `setup.py` script so that it takes `dronecode_sdk/bin/backend_bin` as a resource. Then I added a `start_mavlink()` function that runs this binary in the background (and kills it when the main process dies, unless the main process is killed with e.g. `SIGKILL`).

The server can be run with something like:

```python
from dronecode_sdk import start_mavlink
start_mavlink()
```

I make the difference between python 3.7+ and python <= 3.7 by using respectively `importlib.resources` and `importlib_resources`.

@xvzf: I don't know how much of a hack this is, so I'd love to get your insights!

This is related to #20.

### DIFF
--- a/dronecode_sdk/__init__.py
+++ b/dronecode_sdk/__init__.py
@@ -51,3 +51,33 @@ def connect(*args, **kwargs):
         globals()[plugin](plugin_manager)
 
     return plugin_manager
+
+
+import atexit
+import os
+import subprocess
+import sys
+
+from dronecode_sdk import bin
+from time import sleep
+
+if sys.version_info >= (3,7):
+    from importlib.resources import path
+else:
+    from importlib_resources import path
+
+def start_mavlink():
+    """
+    Starts the gRPC server in a subprocess, listening on localhost:50051
+    """
+    with path(bin, 'backend_bin') as backend:
+        p = subprocess.Popen(os.fspath(backend), shell=False, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
+
+    # Wait to make sure the gRPC server is started.
+    # There are better, yet more complex ways to do that.
+    sleep(2)
+
+    def cleanup():
+        p.kill()
+
+    atexit.register(cleanup)
--- a/setup.py
+++ b/setup.py
@@ -37,12 +37,13 @@ def parse_long_description():
 
 setup(
     name="dronecode-sdk",
-    version="0.1.0a",
-    description="Pythonwrapper for Dronecode SDK",
+    version="0.1.0",
+    description="Python wrapper for Dronecode SDK",
     long_description=parse_long_description(),
     url="https://github.com/dronecode/DronecodeSDK-Python",
-    author="@TODO",
-    author_email="@TODO",
+    maintainer="Jonas Vautherin, Julian Oes",
+    maintainer_email="jonas@auterion.com, julian.oes@auterion.com",
+    include_package_data=True,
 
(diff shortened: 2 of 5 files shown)


dronecode_sdk/__init__.py
```diff
@@ -51,3 +51,31 @@ def connect(*args, **kwargs):
...
+    from importlib_resources import path
+
+def start_mavlink():
+    """
+    Starts the gRPC server in a subprocess, listening on localhost:50051
+    """
+    with path(bin, 'backend_bin') as backend:
+        p = subprocess.Popen(os.fspath(backend), shell=False, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
+
+    sleep(2)
```
julianoes: Why this?

dronecode_sdk/__init__.py
```diff
@@ -51,3 +51,31 @@ def connect(*args, **kwargs):
...
+
+    def cleanup():
+        p.kill()
+
+    atexit.register(cleanup)
```
julianoes: Nice.

setup.py
```diff
@@ -43,6 +43,7 @@ def parse_long_description():
     url="https://github.com/dronecode/DronecodeSDK-Python",
     author="@TODO",
     author_email="@TODO",
```
julianoes: You and or? Or @xvzf :smile:

dronecode_sdk/__init__.py
JonasVautherin: :sweat_smile: because I want to make sure the server is started before `start_mavlink()` returns. In Swift we use a library instead of running an executable in a subprocess, so we get a callback when the server is ready. Here we don't.

Improvements could be one of:
* use the library instead of the executable, but it implies wrapping the C++ interface (with Cython or something like that?)
* check the grpc server status instead of a dumb `sleep`

julianoes: Ok, we can improve it later. For now I'd add a comment describing why.

dronecode_sdk/__init__.py
```diff
@@ -51,3 +51,33 @@ def connect(*args, **kwargs):
...
+
+from dronecode_sdk import bin
+from time import sleep
+
+if sys.version_info >= (3,7):
+    from importlib.resources import path
+else:
+    from importlib_resources import path
+
+def start_mavlink():
```
xvzf: This should work, but somehow we have to make sure we kill the process when python exits (due to a failure or sth else). It _should_ but it happens sometimes that it can't manage to cancel it

JonasVautherin: You mean `atexit` will not always kill it? I believe there is no way to make sure of that. It seems to work well when the program stops normally or raises an error. If you SIGKILL, then there is no way (the Python code cannot intercept that, otherwise it could ignore it and that's not the point of SIGKILL).

I can't find a better way to run the server in its own process without ever having a chance to leak. We could run it as a thread, but that requires to wrap `libbackend.so` (maybe with Cython?).

xvzf: That's correct! I was thinking about a randomized port for the GRPC service.

Cython might get more complicated, I'd suggest to stick with as few dependencies as possible!

JonasVautherin: Good point for the randomized port. Let's keep that in mind as an enhancement for now. But I agree it should be done!

PR #63, opened 2019-05-06
Update proto submodule with offboard proto def and core plugin list.

cswkim:
There are a number of commits between the one for adding `offboard.proto` and the current commit hash (`a54b353d73`), need someone to verify it is OK to include them.

### DIFF
--- a/dronecode_sdk/__init__.py
+++ b/dronecode_sdk/__init__.py
@@ -18,6 +18,7 @@ CORE_PLUGINS = [
     "Gimbal",
     "Info",
     "Mission",
+    "Offboard",
     "Telemetry"
     ]
 


JonasVautherin:
I'll check that tomorrow! It should be fine, but I'd like to run the examples to make sure nothing is broken.

Did you test that offboard is working from python with this update?

cswkim:
So far so good! I'm writing an example file, I can set an initial setpoint, start, move north at a certain rate. I'm trying out more parts, looking at the [guide](https://sdk.dronecode.org/en/guide/offboard.html) and seeing if I can replicate some of those examples. For me, I need to understand the "business rules" of offboard mode so that I can properly test out the feature. What happens if I do X and the expected result is Y but I get Z? Am I doing something wrong or is the plugin code broken?

I'm still trying to get to a point where I can say, "OK I want the drone to go up 1m, then turn right and go 2m, then turn at an angle 40 degrees and go 5m, then stop and hold/loiter" and be able to translate that into offboard api calls.

JonasVautherin:
Awesome! So this will work with [backend 0.16.0+](https://github.com/Dronecode/DronecodeSDK/releases/tag/v0.16.0).

PR #66, opened 2019-05-15
Add offboard example

cswkim:
I also added a little disclaimer about a couple of items, something I noticed when trying to use the `takeoff_and_land.py` example, pointed out by @julianoes. I'm not sure if the comments are worded correctly or need more context. I believe it's because LOITER and HOLD modes require some sort of home position for the RTL fail-safe, which is not available in non-gps set-ups?

### DIFF
--- a/examples/offboard.py
+++ b/examples/offboard.py
@@ -1,0 +1,215 @@
+#!/usr/bin/env python3
+
+"""
+Some caveats when attempting to run the examples in non-gps environments:
+
+- `drone.action.arm()` will return a `COMMAND_DENIED` result because the action requires switching
+  to LOITER mode first, something that is currently not supported in a non-gps environment. You will
+  need to temporarily disable this part here:
+  `https://github.com/Dronecode/DronecodeSDK/blob/develop/plugins/action/action_impl.cpp#L61-L65`
+
+- `drone.offboard.stop()` will also return a `COMMAND_DENIED` result because it requires a mode
+  switch to HOLD, something that is currently not supported in a non-gps environment.
+"""
+
+import asyncio
+
+from dronecode_sdk import connect as dronecode_sdk_connect
+from dronecode_sdk import (
+    Attitude,
+    OffboardError,
+    PositionNEDYaw,
+    VelocityBodyYawspeed,
+    VelocityNEDYaw,
+)
+
+drone = dronecode_sdk_connect(host="127.0.0.1")
+
+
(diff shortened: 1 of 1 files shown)


examples/offboard.py
```diff
@@ -0,0 +1,215 @@
+#!/usr/bin/env python3
+
+"""
+Some caveats when attempting to run the examples in non-gps environments:
+
+- `drone.action.arm()` will return a `COMMAND_DENIED` result because the action requires switching
+  to LOITER mode first, something that is currently not supported in a non-gps environment. You will
+  need to temporarily disable this part here:
+  `https://github.com/Dronecode/DronecodeSDK/blob/develop/plugins/action/action_impl.cpp#L61-L65`
```
julianoes: Good note, and we should find an easier workaround.

JonasVautherin:
Very nice, thanks a lot!

@julianoes: I'm not sure about the discovery thing. None of the examples do it anyway, so I'd say it is not in the scope of this specific PR.

My opinion is that I like keeping the examples to the point. We can write a discovery example, but it doesn't mean that all the examples need to do that, just like we don't check for the telemetry health or, say, "in air" status before arming. And we don't `try... except` everywhere either. But a "real" app would do all that.

This said, there is maybe some work to do to improve the error message when no drone has been discovered.

cswkim:
I would like to write a more robust or "proper" implementation for python, similar to some of the `C++` examples which do connection, health and in-air checks with more error handling. For me personally, if I were brand new to the python sdk, it would be very helpful to see this kind of example, maybe a best practice guide. Is there a language-agnostic approach to general usage? OK first, check connection state, then do a telemetry health all ok check, then try to arm and error handle, etc, etc.

I could beef up this offboard example with all of the above and we could have a running discussion on what best practice might be for the python sdk. Or maybe include sample code in the README after the instructions on how to set-up?

JonasVautherin:
Let's merge that already, and we can always improve the examples later :-).

@cswkim @julianoes: I understand your points about having more verbose examples. I don't have a super strong opinion, actually. Either we can make a difference between "simple/incomplete" examples and make one or two bigger ones (that would check all the status and run a mission for instance), or we can add all the checks in all the tests, so that there is no way to miss them :D.

julianoes:
> or we can add all the checks in all the tests, so that there is no way to miss them :D.

I think there is no way around that.

PR #68, opened 2019-05-24
dcsdkgen got renamed to protoc-gen-dcsdk

JonasVautherin:
@unipheas: this should fix your build issue, but you may have to reinstall the plugin.

I tested in a venv with:

```sh
# Clone the repo
git clone https://github.com/dronecode/dronecodesdk-python
cd dronecodesdk-python

# Create and activate a venv
python -m venv venv
source ./venv/bin/activate

# Install the protoc plugin, now called `protoc-gen-dcsdk`
cd proto/pb_plugins
pip install -r requirements.txt
pip install -e .

# Go back to root of repo and generate the SDK
cd ../..
pip install -r requirements.txt -r requirements-dev.txt
sh other/tools/run_protoc.h

# Optional: check that the protoc plugin is correctly installed in the venv:
which protoc-gen-dcsdk
```

### DIFF
--- a/other/tools/run_protoc.sh
+++ b/other/tools/run_protoc.sh
@@ -34,7 +34,7 @@ function generate {
 
         # Generate plugin
         python3 -m grpc_tools.protoc -I$(dirname ${PROTO_FILE}) \
-                                     --plugin=protoc-gen-custom=$(which dcsdkgen) \
+                                     --plugin=protoc-gen-custom=$(which protoc-gen-dcsdk) \
                                      --custom_out=${PLUGIN_DIR} \
                                      --custom_opt=file_ext=py \
                                      ${PROTO_FILE}


unipheas:
Per these instructions I had to use `git clone https://github.com/dronecode/dronecodesdk-python --recursive` to download the proto/pb_plugins folder. I also ran `sh other/tools/run_protoc.sh` rather than run_protoc.h. Everything was generated successfully and `which protoc-gen-dcsdk` returns `/Users/myuser/Desktop/dronecodesdk-python/venv/bin/protoc-gen-dcsdk`. However, when I go to run `examples/takeoff_and_land.py` I'm left with 1 error now instead of the multiples before. This is the error: 

```
    File "examples/takeoff_and_land.py", line 5, in <module>
        from dronecode_sdk import connect as dronecode_sdk_connect
ModuleNotFoundError: No module named 'dronecode_sdk'
```

The simulator jMavsim is running and the backend is connected.

JonasVautherin:
Oh right, because at the end you need to install the SDK in order to use it from the example.

So from the root of the repo, run `pip install -e .`. I'm updating the instructions above.

unipheas:
I'm now getting this error after running `example/takeoff_and_land.py`
```
  File "examples/takeoff_and_land.py", line 5, in <module>
    from dronecode_sdk import connect as dronecode_sdk_connect
  File "/Users/myuser/Desktop/dronecodesdk-python/dronecode_sdk/__init__.py", line 48, in <module>
    from .plugins import * # NOQA
  File "/Users/myuser/Desktop/dronecodesdk-python/dronecode_sdk/plugins/__init__.py", line 1, in <module>
    -e # -*- coding: utf-8 -*-
NameError: name 'e' is not defined
```

[Here](https://gist.github.com/unipheas/4ea0d141fd8ea0341806470b014bc68d) is the entire log for everything I'm doing, hope this helps

JonasVautherin:
Let's track that in a new issue (#69).

PR #72, opened 2019-05-31
dronecode_sdk: set development status and programming language

JonasVautherin:
- I realized it has to be "Production" OR "Alpha" :laughing:.
- We don't support all of Python 3, so my hope is that setting Python 3.6 means that it works for Python 3.6+. @xvzf do you know about that?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -46,11 +46,10 @@ setup(
     include_package_data=True,
 
     classifiers=[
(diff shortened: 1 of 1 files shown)


xvzf:
@JonasVautherin If we support 3.6 it should also work on 3.6+, as long as it is python 3!

PR #73, opened 2019-05-31
readme: update with the new way of building the backend

JonasVautherin:
@unipheas: I integrated some of your suggestions. Feel free to open a PR in this repo if you think something is unclear!
@xvzf: FYI, the backend now uses cmake only


unipheas:
All works on my system starting from scratch!!

Issue #81, opened 2019-06-30
Offboard Mode Problem (status = StatusCode.UNIMPLEMENTED)


bozkurthan:
Hello I' tried to switch offboard mode from examples/offboard.py by running `examples/offboard.py` in PX4 Sitl gazebo v1.9.0.

But I got following error:

```
bozkurthan@bozkurthan:~/Desktop/MAVSDK-Python-master$ python3 examples/offboard.py
-- Arming
-- Setting initial setpoint
Traceback (most recent call last):
File "examples/offboard.py", line 217, in <module>
loop.run_until_complete(run_offb_ctrl_velocity_ned())
File "/usr/lib/python3.6/asyncio/base_events.py", line 484, in run_until_complete
return future.result()
File "examples/offboard.py", line 38, in run_offb_ctrl_velocity_ned
await drone.offboard.set_velocity_ned(VelocityNEDYaw(0.0, 0.0, 0.0, 0.0))
File "/home/bozkurthan/.local/lib/python3.6/site-packages/mavsdk/plugins/offboard.py", line 704, in set_velocity_ned
response = await self._stub.SetVelocityNed(request)
File "/home/bozkurthan/.local/lib/python3.6/site-packages/aiogrpc/channel.py", line 40, in __call__
return await fut
grpc._channel._Rendezvous: <_Rendezvous of RPC that terminated with:
        status = StatusCode.UNIMPLEMENTED
        details = ""
        debug_error_string = "{"created":"@1561876519.775117151","description":"Error received from peer ipv4:127.0.0.1:50051","file":"src/core/lib/surface/call.cc","file_line":1046,"grpc_message":"","grpc_status":12}"
>
```

I tried to run with `python3 examples/offboard.py ` as well but same error appeared. examples/takeoff_and_land.py worked without error.
By the way, I installed MAVSDK with pip3.

Thanks for helping

julianoes:
@JonasVautherin do you know what's happening here?

bozkurthan:
The error still exists.. FYI @irsdkv

bozkurthan:
```
-- Arming
-- Setting initial setpoint
Traceback (most recent call last):
  File "examples/offboard.py", line 217, in <module>
    loop.run_until_complete(run_offb_ctrl_velocity_ned())
  File "/usr/lib/python3.6/asyncio/base_events.py", line 484, in run_until_complete
    return future.result()
[...]

PR #83, opened 2019-07-02
Proto submodule update

irsdkv:
Update Proto submodule revision due to fix https://github.com/mavlink/MAVSDK-Python/issues/81


JonasVautherin:
Thanks @irsdkv!

julianoes:
@JonasVautherin we probably need to make a new release with this?

JonasVautherin:
Hmm 0.1.0 does work, this only fixes the `master` build (when people try to run it against the latest `mavsdk_server`). 0.1.0 just runs `backend_bin`, but that's transparent to the user.

0.1.0 does not lack features, I believe. I don't think it is urgent to release. I need to write scripts to make the release process more automated, and I will release at that point (if that's okay for you).

PR #89, opened 2019-07-17
Update __init__.py to reflect geofence proto

coderkalyan:
Here is an updated `__init__.py` reflecting the new Geofence gRPC binding I have been working on. For this PR to be of any use, https://github.com/mavlink/MAVSDK-Proto/pull/89 must be merged first so the `.proto` files are introduced.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "proto"]
 	path = proto
-	url = https://github.com/Dronecode/DronecodeSDK-Proto.git
+	url = https://github.com/mavlink/MAVSDK-Proto.git
--- a/mavsdk/__init__.py
+++ b/mavsdk/__init__.py
@@ -16,6 +16,7 @@ CORE_PLUGINS = [
     "Calibration",
     "Camera",
     "Core",
+    "Geofence",
     "Gimbal",
     "Info",
     "Mission",


coderkalyan:
@JonasVautherin no problem :) Sorry, I didn't understand exactly what you want me to do to the submodule?

JonasVautherin:
So once your `geofence.proto` PR is merged, you'll be able to go in `proto/` in MAVSDK-Python, which is a git submodule pointing to MAVSDK-Proto. And there you can just `git checkout master && git pull`, or something equivalent, and add that change to this PR.

Because this PR is useless if the `proto/` submodule does not contain the `geofence.proto` file :-).

Does that help?

coderkalyan:
Yep! Let me do that now

coderkalyan:
@JonasVautherin I have updated the submodules. I also took the chance to update the submodule remote from DronecodeSDK/DronecodeSDK-Proto to mavlink/MAVSDK-Proto (I know the first one redirects). If you don't want this I can remove it.

JonasVautherin:
Seems good! I'd like to merge that after https://github.com/mavlink/MAVSDK/pull/805, otherwise it will just throw `UNIMPLEMENTED` and that will be confusing.

PR #90, opened 2019-07-18
Syntax error printf

bozkurthan:
Small correction for takeoff and land example.


JonasVautherin:
Not sure I get that. Why is that required? Doesn't it work on your side?

bozkurthan:
It my false python version is conflicted when I call example (3.5 instead of 3.6) so I was getting syntax error. I'm not expert on 3.6 :) I'm closing.

PR #104, opened 2019-08-30
Documentation templates

### LINKED ISSUE #82: Generate reference docs for Python frontend
@JonasVautherin we should generate reference docs for the Python bindings.

JonasVautherin:
Update the templates to generate documentation from the proto files.

Note: the second commit is the interesting one.

Fixes #82.


other/templates/request.j2
```diff
@@ -1,8 +1,27 @@
 
 async def {{ name.lower_snake_case }}(self{% for param in params %}, {{ param.name.lower_snake_case }}{% endfor %}):
-    """ Generated by dcsdkgen
+    """
```
julianoes: According to https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings there is no newline after `"""`. What's your take on that?

other/templates/request.j2
```diff
@@ -1,8 +1,27 @@
...
+ {{ indent(method_description, 1) }}
+
+     {% if params -%}
+     Parameters
+     ----------
```
julianoes: According to PEP it would probably be more like this:
```
    Parameters:
```

JonasVautherin: I followed the [numpy](https://numpydoc.readthedocs.io/en/latest/format.html) examples, I really don't mind. I just know that PyCharm understands what we have now. I will need to check what happens with `Parameters:`.

I don't find where in PEP it defines how documentation should be written. If there is something that is more standard than numpy, I'm fine changing it.

other/templates/request.j2
JonasVautherin: I don't care :-). I'll change it.

other/templates/request.j2
julianoes: https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

That's all I found. But otherwise go with what numpy suggests, looks fine.

other/templates/request.j2
JonasVautherin: Oh, I see why it is like this. If I don't do that, then the next paragraph is not aligned correctly, and that results in a Python indentation error.

Struggling a bit with jinja2 sometimes :sweat_smile:

PR #105, opened 2019-08-30
Add drone class

### LINKED ISSUE #53: Abstract away gRPC errors
As reported in #49, the way we handle gRPC errors is not elegant (and leaking gRPC to the user.

For instance, this is the output if the backend is not reachable:

```
[ ERROR] <_Rendezvous of RPC that terminated with:
        status = StatusCode.UNAVAILABLE
        details = "Name resolution failure"
        debug_error_string = "{"created":"@1550220055.531945000","description":"Failed to create subchannel","file":"src/core/ext/filters/client_channel/client_channel.cc","file_line":2261,"referenced_errors":[{"created":"@1550220055.531943000","description":"Name resolution failure","file":"src/core/ext/filters/client_channel/request_routing.cc","file_line":163,"grpc_status":14}]}"
>
File "/Users/docs/boulot/controleur/controleur/Autopilot.py", line 68, in connect
autopilot_type = await Autopilot.vehicle.info.get_version()
File "/Users/docs/boulot/controleur/controleur/Libs/DronecodeSDK-Python/dronecode_sdk/plugins/info.py", line 288, in get_version
response = await self._stub.GetVersion(request)
File "/Users/.local/share/virtualenvs/controleur-7At54XPt/lib/python3.6/site-packages/aiogrpc/channel.py", line 40, in __call__
return await fut
grpc._channel._Rendezvous: <_Rendezvous of RPC that terminated with:
        status = StatusCode.UNAVAILABLE
        details = "Name resolution failure"
        debug_error_string = "{"created":"@1550220055.531945000","description":"Failed to create subchannel","file":"src/core/ext/filters/client_channel/client_channel.cc","file_line":2261,"referenced_errors":[{"created":"@1550220055.531943000","description":"Name resolution failure","file":"src/core/ext/filters/client_channel/request_routing.cc","file_line":163,"grpc_status":14}]}"
>
```

As mentioned by @zulufoxtrot, the error can be handled with something like:

```
except grpc.RpcError as e:
logger.exception("Exception in telemetry handler: ")

if e._state.code == grpc.StatusCode.UNAVAILABLE:
logger.error("DronecodeSDK backend is unreacheable.")
```

We should probably handle that inside the SDK and raise a proper exception when that happens.

### LINKED ISSUE #86: Document `mavsdk_connect` vs. `start_mavlink`
I often see people confused about the usage of `mavsdk_connect` vs. `start_mavlink`. We should either make the API clearer or document it better, I think.

JonasVautherin:
This creates a `Drone` class, for different reasons:

* The previous implementation was registering the plugins at runtime, and therefore the auto-completion (e.g. in PyCharm) was mostly inexistant. Now the auto-completion works.
* I enables a new syntax that (I believe) is clearer.
    * If you don't know what you are doing, `drone = Drone()` and `drone.connect()` do what you expect.
    * If you forget to `connect()`, you will get a error saying "did you call `Drone.connect()`?" when you try to call a plugin.
* Bonus: we now wait for `mavsdk_server` before returning on `drone.connect()`. Previously, we were getting a gRPC `UNAVAILABLE` error that was confusing to users.

Fixes #86, fixes #53.


mavsdk/drone.py
```diff
@@ -0,0 +1,129 @@
...
+        return self._plugins["offboard"]
+
+    @property
+    def telemetry(self) -> Telemetry:
+        if "telemetry" not in self._plugins:
+            raise RuntimeError("Telemetry plugin has not been initialized! Did you run `Drone.connect()`?")
+        return self._plugins["telemetry"]
+
+    @staticmethod
+    def _start_mavsdk_server(drone_address=None):
```
julianoes: Why `drone_address`? What about `mavlink_address`?

examples/offboard_attitude.py
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/env python3
+
+"""
+Some caveats when attempting to run the examples in non-gps environments:
```
julianoes:
```suggestion
Some caveats when attempting to run the examples in non-GPS environments:
```

everywhere...

examples/telemetry_takeoff_and_land.py
```diff
@@ -1,15 +1,38 @@
...
+    Finally, it goes to the actual works: arm the drone, initiate a takeoff
+    and finally land.
+
+    Note that "observe_is_in_air" is not necessary, but it ensures that the
+    script waits until the drone is actually landed, so that we receive feedback
+    during the landing as well.
+    """
+
+    # Init the drone
+    drone = Drone()
```
julianoes: I'm not sold on using the word `drone` because it can also be a plane or a rover or whatnot. On the C++ side this is called a `system`.

We don't have to change that in this PR but I would like to have it consistent at some point. We can also consider a names like `vehicle` or `robot`.

mavsdk/drone.py
JonasVautherin: That was suggested in #86. I really don't mind. Should I go for `mavlink_address`? @hamishwillee any opinion?

examples/telemetry_takeoff_and_land.py
JonasVautherin: I don't really mind, but I feel like `drone` is used generically today: e.g. "the SpaceX droneship", and I don't personally associate `drone` with `multicopter`. As a developer, I am super fine writing `rover = Drone()`, too.

I feel like `drone` or `vehicle` are describing the kind of object MAVSDK usually controls, whereas `system` is so generic it doesn't mean anything, while at the same time leaking MAVLink terminology. I don't care about whether or not it is abusive to call a ship "a drone" or "a vehicle", as long as people get what it is.

Anyway, I would also be fine with system, because anyway that's only used once in the whole code (`drone = System()`, or `rover = Vehicle()`, or `ship = Drone()`). I just would like to settle on something because it is starting to be time-consuming to rename those kinds of things in all the MAVSDK projects :blush:.

julianoes: I agree, I'm just bringing it up for consistency. Given it's the MAV SDK it should be MAV :smile:, or Mav :thinking:.

However, it's also a system, so it can be a gimbal, camera, drone, rover, plane, ...

mavsdk/drone.py
JonasVautherin: It's now `system_address`

examples/telemetry_takeoff_and_land.py
JonasVautherin: `System` it is :smile:

PR #111, opened 2019-09-28
Mocap plugin

irsdkv:
Due to https://github.com/mavlink/MAVSDK-Proto/pull/105


JonasVautherin:
I think this should now be updated with the latest `MAVSDK-Proto` version :+1:.

irsdkv:
Should I do anything else?
`Proto` PR was merged and seems like current implementation works ok

PR #123, opened 2019-11-01
Add mavsdk_server on Windows

julianoes:
This seemed to work on my 64bit Windows in Python 3.7 64bit and Python 3.7 32bit.

Note that the gRPC dependencies is not pre-built for Python 3.8 yet.

@JonasVautherin please check if the README is still correct. I believe it needs updating and wasn't really sure which steps are required and which are not.


README.md
```diff
@@ -72,6 +72,7 @@ Run the following helper script. It will generate the Python wrappers for each p
 After generating the wrapper you can install a development (editable) version of the package using:
 
 ```
+python3 setup.py build
```
JonasVautherin: Why is that required? So that it downloads `mavsdk_server` from github? I guess we should add more details to explain that one may want to use `mavsdk_server` built from sources instead?

I personally never do `python setup.py build`: either I do `pip install mavsdk`, or I build everything from sources.

julianoes: Ok, that's fine. I just didn't know how to test `pip install`, so that's what enabled me to do that.

PR #129, opened 2019-11-07
Add travis-ci

julianoes:
Best case we can eventually automatically deploy to PyPi on all platforms.


JonasVautherin:
Would be awesome.

PR #141, opened 2019-12-10
travis-ci: add Windows 32 bit

julianoes:
It's sad that this is needed.

The problem seems to be that we can't upload the same file twice to PyPi just with another name.


julianoes:
Don't worry, we now run the `setup.py` script in Python 32 bit and 64 bit and upload both. The compiled binary is 32 bit for both which seems to work fine.

PR #150, opened 2019-12-19
Update Shell plugin

irsdkv:
If case if that addition will be merged in MAVSDK (https://github.com/mavlink/MAVSDK/pull/944)


irsdkv:
Cancelled due to [this](https://github.com/mavlink/MAVSDK-Proto/pull/112) conversation

PR #171, opened 2020-04-03
C++ auto-generation ripple effects

julianoes:
Work in progress. For now it contains action and telemetry changes.


mavsdk/generated/action.py
```diff
@@ -322,9 +322,34 @@ def _extract_result(self, response):
             raise ActionError(result, "reboot()")
         
 
+    async def shutdown(self):
+        """
+         *
```
JonasVautherin, line 327: I don't get where this `*` comes from :thinking:

JonasVautherin: Oh, in the proto file it starts with `/**` instead of `/*`. Fixing that!

JonasVautherin: Fixed in https://github.com/mavlink/MAVSDK-Proto/pull/121

JonasVautherin:
So the plan here is to wait until all the plugins are generated in C++ and then merge that, or would you like to merge it before?

julianoes:
Yes, waiting is fine.

JonasVautherin:
I updated the protos to be on the latest state of the `proto` submodule

JonasVautherin:
@julianoes: I think it is ready now. For some reason, terminating the async tasks in the examples now gives an error:

```an error occurred during closing of asynchronous generator <async_generator object Mission.mission_progress at 0x7f700e4889d0>
asyncgen: <async_generator object Mission.mission_progress at 0x7f700e4889d0>
RuntimeError: aclose(): asynchronous generator is already running
```

But I'm not sure exactly what this is about. I think it's not in the scope of this PR.

PR #183, opened 2020-05-11
Fix tasks termination for Python 3.8

JonasVautherin:
See discussion here: https://bugs.python.org/issue38559

Before calling `aclose()` (which is called by `shutdown_asyncgens()`), Python 3.8 requires the tasks to be cancelled.


JonasVautherin:
CI is failing because the Windows binary is not released in MAVSDK, that's expected.

PR #195, opened 2020-05-30
Generate API docs

### LINKED ISSUE #82: Generate reference docs for Python frontend

julianoes:
I had some fun (or not) with sphinx.

@hamishwillee I'm open to generate markdown or xml instead and integrate it into the existing docs but this is better than nothing for now.
http://mavsdk-python-docs.s3-website.eu-central-1.amazonaws.com/

Fixes #82, completely.

Next step is to do it in CI.


hamishwillee:
@julianoes 
1. FWIW I am OK with Sphinx for these docs, and more generally that docs for the different systems being in the format expected by those systems. Great job.

What I would do is:
- Move the stuff in the github repo README into a PAGE in https://mavsdk.mavlink.io/develop/en/index.html - rather than linking to the README.
- That page would have the links to the readme, examples in the github repo, using pip, api reference
- The API reference would link back to this page on Python in the main docs.

So the official docs are the "hub" rather than the Python repo. We can then choose to move things around later if we want.

julianoes:
@hamishwillee I agree that we could move that over but the disadvantage is always that things can get out of sync. In this case even more so because mavsdk.mavlink.io is versioned with the C++ repo and not the Python repo.

hamishwillee:
Good point. I still think this would be better centralised outside of the readme - presumably in the Sphinx docs. Not my call.

julianoes:
Right. I have copied the simpler usage part to the Sphinx docs but left the more involved things like building yourself in the readme.

PR #207, opened 2020-06-03
Make modules out of the plugins

### LINKED ISSUE #200: mission.py example broken (type object 'MissionItem' has no attribute 'CameraAction')
Trying to execute the mission.py example I get:
```
Waiting for mavsdk_server to be ready...
Connected to mavsdk_server!
Waiting for drone to connect...
Drone discovered with UUID: 255
  File "mission.py", line 103, in <module>
    loop.run_until_complete(run())
  File "/usr/lib/python3.8/asyncio/base_events.py", line 616, in run_until_complete
    return future.result()
  File "mission.py", line 32, in run
    MissionItem.CameraAction.NONE,
AttributeError: type object 'MissionItem' has no attribute 'CameraAction'
```
I was able to reproduce this in a clean environment - Dockerfile:
```
FROM ubuntu:20.04
RUN apt-get update && apt-get install -y python3-pip git vim
RUN pip3 install mavsdk
```

JonasVautherin:
An issue we have in the current system (where all the files are generated in `mavsdk/generated`) is that we don't have any namespacing between them. Therefore, `from mavsdk import MissionItem` could import `MissionItem` from `Mission`, or from `MissionRaw`. In this case, it broke the `examples/mission.py` example.

I therefore made a module out of each plugin, so that `import mavsdk.mission` is a thing, and one could `from mavsdk.mission_raw import MissionItem as MissionRawItem`, if needed.

I had to make a few changes for that:
* Remove the `globals()` use in `system.py`, mainly because I did not really understand how it worked
* Generate the files directly in `mavsdk/`, because I could not make the modules thing work in `mavsdk/generated/` without ending up with imports of the form `from mavsdk.generated.mission import ...`. Maybe there is a way, I just did not find it.
* Update the scripts here and there to generate files accordingly
* Update the examples

I tested the examples and it seems to be working for me. I would be glad if @petergerten could confirm that it fixes the mission example :blush:.

I did not test against the docs generation (mainly because it's getting late), but I wanted @julianoes' opinion first.

Note: I tried to make commits that make sense for the review, so the auto-generation is isolated in two commits :innocent:.

Fixes #200


examples/camera.py
```diff
@@ -2,7 +2,7 @@
 
 import asyncio
 
-from mavsdk import (CameraError, Mode)
+from mavsdk.camera import (CameraError, Mode)
```
julianoes, line 5: I would be tempted to just do:
```
from mavsdk import camera
```
and then access it using:
```
camera.Mode
```

That's possible, right?

JonasVautherin: Yes that's possible, too. Or just `import mavsdk` and then use `mavsdk.camera.Mode`. I think it's really up to the user :+1:.

JonasVautherin:
Thanks for the review! I'll try to look into the documentation side sometime next week, because I probably broke it. In the meantime  I hope that @petergerten can work from this branch for the missions.

JonasVautherin:
Also in the code? In my testing, in the code I could do `mavsdk.action.ActionResult`... Maybe the docs needs a path and not a module (because it's `mavsdk/action/action.py`)

julianoes:
```
import mavsdk
```

```
help(mavsdk.action):

Help on package mavsdk.action in mavsdk:

NAME
    mavsdk.action

PACKAGE CONTENTS
    action
    action_pb2
    action_pb2_grpc

FILE
    /home/julianoes/src/MAVSDK-Python/mavsdk/action/__init__.py
```

```
help(mavsdk.action.action)

Help on module mavsdk.action.action in mavsdk.action:

NAME
    mavsdk.action.action - # -*- coding: utf-8 -*-

CLASSES
    builtins.Exception(builtins.BaseException)
        ActionError
    builtins.object
        ActionResult
    mavsdk._base.AsyncBase(builtins.object)
        Action
    
    class Action(mavsdk._base.AsyncBase)
     |  Action(async_plugin_manager)
     |  
     |  Enable simple actions such as arming, taking off, and landing.
     |  
     |  Generated by dcsdkgen - MAVSDK Action API
     |  
     |  Method resolution order:
     |      Action
     |      mavsdk._base.AsyncBase
     |      builtins.object
     |  
     |  Methods defined here:
     |  
     |  async arm(self)
     |      Send command to arm the drone.
     |      
     |      Arming a drone normally causes motors to spin at idle.
     |      Before arming take all safety precautions and stand clear of the drone!
     |      
     |      Raises
     |      ------
     |      ActionError
     |          If the request fails. The error contains the reason for the failure.
...
```

```
help(mavsdk.action.Action):

Help on class Action in module mavsdk.action.action:

class Action(mavsdk._base.AsyncBase)
 |  Action(async_plugin_manager)
 |  
 |  Enable simple actions such as arming, taking off, and landing.
 |  
 |  Generated by dcsdkgen - MAVSDK Action API
 |  
 |  Method resolution order:
 |      Action
 |      mavsdk._base.AsyncBase
 |      builtins.object
 |  
 |  Methods defined here:
 |  
 |  async arm(self)
 |      Send command to arm the drone.
 |      
 |      Arming a drone normally causes motors to spin at idle.
 |      Before arming take all safety precautions and stand clear of the drone!
 |      
 |      Raises
 |      ------
 |      ActionError
 |          If the request fails. The error contains the reason for the failure.
...

```

JonasVautherin:
Doesn't seem wrong to me. Is it?

`mavsdk.action.Action` is the class, and `mavsdk.action.action` is the `action.py` file.

PR #213, opened 2020-06-08
Fix misspelling in mission_raw.py

Nicole426:


Nicole426:
I can do it

Nicole426:
At least I thought I could. Can not manage to do the generation. So it might be easier, if you can do it quickly. I will try to get it work nevertheless.

Sorry, that I haven't seen, that the changes have to be in the templates earlier.

JonasVautherin:
Maybe you can try with a venv (feel free to adapt those commands if you want):

```
cd /path/to/mavsdk-python
python3 -m venv venv
./venv/bin/activate
pip install -r requirements-dev.txt -r requirements.txt
cd proto/pb_plugins
pip install -r requirements.txt
pip install -e .
cd ../.. # back to /path/to/mavsdk-python, the root of this repo
./other/tools/run_protoc.sh
```

What it does is this:
1. install the dependencies (requirements.txt) of MAVSDK-Python
2. install the dependencies of `protoc-gen-dcsdk`
3. install `protoc-gen-dcsdk`
4. run the autogen script, that will use `protoc-gen-dcsdk`

If this doesn't work, I'll do it, but I think it's nice if you can do it yourself :blush:

Nicole426:
Thanks for your input. I tried it with the venv as well, but the result was the same. 
The installation of the dependencies and proto-gen-dcsdk worked, but when I'm running the autogen script, a new windows open shortly and closes directly (and does not do anything I think). I'm totally unfamiliar with shell scripts etc., so I don't know, what might be wrong.

JonasVautherin:
Ok, I fixed it in #215. I kept your commit though, because you found the bug so you deserve credit :wink:

PR #235, opened 2020-07-23
examples: closer goto, and go up, not down

julianoes:
This is required because the goto altitude is absolute.

### DIFF
--- a/examples/goto.py
+++ b/examples/goto.py
@@ -28,7 +28,10 @@ async def run():
 
     await asyncio.sleep(1)
 
-    await drone.action.goto_location(55.8688660, -4.2851267, 40, 0)
+    await drone.action.goto_location(47.399386, 8.535245, 500, 0)
+
+    while True:
+        await asyncio.sleep(1)
 
 if __name__ == "__main__":
     loop = asyncio.get_event_loop()
--- a/examples/goto.py
+++ b/examples/goto.py
@@ -28,10 +28,7 @@ async def run():
 
     await asyncio.sleep(1)
 
-    await drone.action.goto_location(47.399386, 8.535245, 500, 0)
-
-    while True:
-        await asyncio.sleep(1)
+    await drone.action.goto_location(47.399386, 8.535245, 500, float('nan'))
 
 if __name__ == "__main__":
     loop = asyncio.get_event_loop()


examples/goto.py
```diff
@@ -28,7 +28,10 @@ async def run():
 
     await asyncio.sleep(1)
 
-    await drone.action.goto_location(55.8688660, -4.2851267, 40, 0)
+    await drone.action.goto_location(47.399386, 8.535245, 500, 0)
+
+    while True:
```
JonasVautherin, line 33: Why is that necessary? :thinking:

iwishiwasaneagle:
@julianoes does this work with `plane`?

examples/goto.py
```diff
@@ -28,7 +28,10 @@ async def run():
 
     await asyncio.sleep(1)
 
-    await drone.action.goto_location(55.8688660, -4.2851267, 40, 0)
+    await drone.action.goto_location(47.399386, 8.535245, 500, 0)
```
iwishiwasaneagle, line 31: Change heading angle to `float('nan')` such that `await drone.action.goto_location(47.399386, 8.535245, 500,float('nan')`

examples/goto.py
julianoes: Oops that was just me debugging.

examples/goto.py
julianoes: Oh yes, that would be nicer.

julianoes:
@iwishiwasaneagle do you know why the plane crashes? Is it just because it's too low for it do to crazy turns?

iwishiwasaneagle:
> @iwishiwasaneagle do you know why the plane crashes? Is it just because it's too low for it do to crazy turns?

Well the docs for (MAV_CMD_DO_REPOSITION)[https://mavlink.io/en/messages/common.html#MAV_CMD_DO_REPOSITION] say that for planes, yaw angle should either be 0 or 1 for loiter direction. However, as soon as I do either of those my SITL Plane loses complete control.  

 > Yaw heading. NaN to use the current system yaw heading mode (e.g. yaw towards next waypoint, yaw to home, etc.). For planes indicates loiter direction (0: clockwise, 1: counter clockwise)

PR #246, opened 2020-08-16
Examples: Added FollowMe example and updated goto.py

TheSharad:
Added an example on how to use the FollowMe plugin and updated the goto.py example with some code that allows user to get the absolute terrain altitude and then use it to fly the drone. This helps in flying the drone irrespective of the location as we can just provide values relative to that altitude.

### DIFF
--- a/examples/follow_me_example.py
+++ b/examples/follow_me_example.py
@@ -35,12 +35,10 @@ async def fly_drone():
     #Arming the drone
     print ("-- Arming")
     await drone.action.arm()
-    await asyncio.sleep(1)
     
     #Follow me Mode requires some configuration to be done before starting the mode
     conf = Config(default_height, follow_distance, direction, responsiveness)
     await drone.follow_me.set_config(conf)
-    await asyncio.sleep(1)
     
     print ("-- Taking Off")
     await drone.action.takeoff()
@@ -64,9 +62,6 @@ async def fly_drone():
     
     print ("-- Landing")
     await drone.action.land()
-    await asyncio.sleep(35)
-    print ("-- Disarming")
-    await drone.action.disarm()
 
 if __name__ == "__main__":
     loop = asyncio.get_event_loop()
--- a/examples/follow_me_example.py
+++ b/examples/follow_me_example.py
@@ -1,0 +1,73 @@
+#! /usr/bin/env python3
+
+#This example shows how to use the follow me plugin
+
+import asyncio
+from mavsdk import System
+from mavsdk.follow_me import (Config, FollowMeError, TargetLocation)
+
+default_height = 8.0 #in Meters
+follow_distance = 2.0 #in Meters, this is the distance that the drone will remain away from Target while following it 
+#Direction relative to the Target 
+#Options are NONE, FRONT, FRONT_LEFT, FRONT_RIGHT, BEHIND
+direction = Config.FollowDirection.BEHIND
+responsiveness =  0.02
+
+#This list contains fake location coordinates (These coordinates are obtained from mission.py example)
+fake_location = [[47.398039859999997,8.5455725400000002],[47.398036222362471,8.5450146439425509],[47.397825620791885,8.5450092830163271]]
+
+async def fly_drone():
+    drone = System()
+    await drone.connect(system_address="udp://:14540")
+    
+    #This waits till a mavlink based drone is connected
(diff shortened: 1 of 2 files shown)


examples/goto.py
```diff
@@ -20,15 +20,21 @@ async def run():
             print("Global position estimate ok")
             break
 
+    print("Fetching Absolute Height of Terrain....")
```
JonasVautherin, line 23: It feels like you want the [`home position`](https://github.com/mavlink/MAVSDK-Proto/blob/master/protos/telemetry/telemetry.proto#L18) here. Would that make sense?

examples/follow_me_example.py
```diff
@@ -0,0 +1,73 @@
...
+    await asyncio.sleep(1)
+    
+    #Follow me Mode requires some configuration to be done before starting the mode
+    conf = Config(default_height, follow_distance, direction, responsiveness)
+    await drone.follow_me.set_config(conf)
+    await asyncio.sleep(1)
+    
+    print ("-- Taking Off")
+    await drone.action.takeoff()
+    await asyncio.sleep(8)
```
JonasVautherin, line 47: Why do we need those `sleep` between actions? :thinking:

examples/goto.py
TheSharad: That is not exactly the home position, actually i was just wanted to get the home altitude (amsl). We can change it to something like "Fetching AMSL", i did this as it makes it easier to use goto_location as we can then just add our required drone altitude to the obtained amsl and provide it to the function.

examples/follow_me_example.py
TheSharad: I guess we do not need it, but i added that there to make sure that drone hovers for some duration after taking off before it actually starts to do something else.. If you feel like we can change it. 
Thanks for your time Jonas..

examples/goto.py
TheSharad: I have changed the print statement to better reflect the reason for fetching the postion() update..

JonasVautherin: What I mean is that the home position gives you the altitude at which you took off. So if it's about having an estimation of the absolute height of the terrain, it seems reasonable to me. It should do the same as what you are doing now, unless the drone is already flying (in which case the home altitude would be what you want).

Or is the home altitude not correct in this case?

examples/follow_me_example.py
JonasVautherin: Right. But I don't think it makes sense to sleep between `set_config` and `takeoff` (see line 43), for instance. Also the one at line 67 (waiting until it lands) makes no sense to me. You could just drop it :+1:.

examples/follow_me_example.py
```diff
@@ -0,0 +1,73 @@
...
+    #Stopping the follow me mode
+    print ("-- Stopping Follow Me Mode")
+    await drone.follow_me.stop()
+    await asyncio.sleep(5)
+    
+    print ("-- Landing")
+    await drone.action.land()
+    await asyncio.sleep(35)
+    print ("-- Disarming")
+    await drone.action.disarm()
```
JonasVautherin, line 69: No need to disarm: if landing succeeds, it will disarm. If it is still in the air after 35 seconds, you probably don't want to disarm (and it will probably be denied anyway) :laughing:

examples/goto.py
TheSharad: Great, I got your point. Actually I misunderstood "position updates" with "home position update"  and yes I was trying to do the same thing and yes it makes more sense to use home position here. Thanks for your suggestion. I will make the changes.

examples/follow_me_example.py
TheSharad: Got the point, I guess I wanted to disarm the drone myself😂 and added that long delay to make sure it lands, Earlier I thought of using the Landed State to check it but then I chose the easier way. Anyways, the drone will automatically disarm after it lands safely😅. Thank you so much for your time & suggestions, I will make the changes now.

TheSharad: Done

examples/goto.py
TheSharad: I have made the changes, please go through them. Thanks.

JonasVautherin (approved):
Thanks! :rocket:

TheSharad:
Thank you so much for your time and suggestions. 😃

PR #252, opened 2020-09-11
Added manual_control Example

GriffinMack:
Added a basic example showcasing the use of the manual control plugin. 

I could not find a way to implement joystick support without importing another pip package, so I decided to show manual input using some hard coded test input vectors. I believe this gets the point across, but the controls are very spastic (though it is fun to watching using jMAVsim).

### DIFF
--- a/examples/manual_control.py
+++ b/examples/manual_control.py
@@ -1,0 +1,84 @@
+#! /usr/bin/env python3
+
+# This example shows how to use the manual controls plugin. Manual inputs are taken from a test set, but the use of a joystick can be implemented using third party extensions.
+
+import asyncio, random
+from mavsdk import System
+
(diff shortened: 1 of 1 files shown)


JonasVautherin (approved):
Yes looks nice! Thanks

julianoes:
Thanks a lot @GriffinMack!

kennethakiti1:
@GriffinMack I have been trying to run the manual_control code on my PX4 drone, however, this doesn't seem to work, the drone just goes up to takeoff heights and then lands immediately. 

if anyone else has tried it, can you give me a recommendation?

kennethakiti1:
@GriffinMack I have been trying to run the manual_control code on my PX4 drone, however, this doesn't seem to work, the drone just goes up to takeoff heights and then lands immediately. 

if anyone else has tried it, can you give me a recommendation?

PR #256, opened 2020-09-15
Updated the Geofence Example

GriffinMack:
Minor changes. The imports were not working correctly before, and I decided to create the boundary around the home point instead of a hard coded location


JonasVautherin (approved):
Thanks a lot!

julianoes (approved):
Thx!

PR #261, opened 2020-10-05
fix protobuf version

NicolasM0:
I had protobuf 3.11.3 which gave me this error when importing mavsdk:
AttributeError: module 'google.protobuf.descriptor' has no attribute '_internal_create_key'

It is ok after updating protobuf to 3.13.0.


JonasVautherin (approved):
Thanks!

PR #283, opened 2020-11-04
Fix docs

julianoes:
This fixes http://mavsdk-python-docs.s3-website.eu-central-1.amazonaws.com/.

Note that the new docs contain the failure plugin which does not work yet.


JonasVautherin (dismissed):
Nice

PR #294, opened 2021-01-04
0.35.0 MAVSDK Server based MAVSDK-Python

hrnbot:
v0.35.0 equivalent MAVSDK-Python


JonasVautherin:
Thanks! I'm fixing the links in https://github.com/mavlink/MAVSDK-Python/pull/295, but that will still fail for armv6-musl. Let me release MAVSDK `v0.35.1`, then you can bump it here to v0.35.1 and it should work together with #295.

Does it make sense? You should be able to do that in a few hours, when the artifacts are published.

JonasVautherin:
Should be fine now if you rebase `master` and use `v0.35.1` :+1:.

JonasVautherin (approved):

PR #310, opened 2021-02-19
Update proto submodule and generate code

JonasVautherin:
Note: this is not adding the new `tracking_server` plugin yet (ideally we would test it first?).


julianoes:
@JonasVautherin I would add the `tracking_server`, that way we can start testing it.

julianoes (approved):

PR #343, opened 2021-05-04
Update MAVSDK_SERVER_VERSION

DonatelloX:
Update to the latest version of mavsdk.


julianoes (commented):
@JonasVautherin does this maybe require an update to the proto submodule as well? I don't quite remember.

DonatelloX:
In my experience the MAVSDK_SERVER_VERSION set to 0.39.0 allows you to use the drone.action.hold()
If I try to leave 0.38.2, hold() doesn't work.

julianoes:
Let me merge this and then quickly update the proto.

PR #370, opened 2021-07-27
Replace print statements with logging statements

qthibeault:
The print statements in the `async_plugin_manager` module can create visual clutter when connecting to multiple systems. Users should be able to decide if those statements are printed by configuring logging on an application level.

By default, if logging is configured the mavsdk connection messages should be printed. Users can easily silence them by selecting the appropriate logger and setting the level to anything above DEBUG.

Example:

```python
logging.getLogger("mavsdk.async_plugin_manager").setLevel(logging.INFO)
```

### DIFF
--- a/mavsdk/async_plugin_manager.py
+++ b/mavsdk/async_plugin_manager.py
@@ -2,8 +2,6 @@
 import logging
 import aiogrpc
 
-_logger = logging.getLogger(__name__)
-
 
 class AsyncPluginManager:
     """
@@ -32,9 +30,13 @@ class AsyncPluginManager:
             "{}:{}".format(self.host, self.port)
         )
 
-        _logger.debug("Waiting for mavsdk_server to be ready...")
+        logger = logging.getLogger(__name__)
+        logger.setLevel(logging.DEBUG)  # Enable debug messages by default
+        logger.addHandler(logging.NullHandler())  # Avoid errors when user has not configured logging
+
+        logger.debug("Waiting for mavsdk_server to be ready...")
         await aiogrpc.channel_ready_future(self._channel)
-        _logger.debug("Connected to mavsdk_server!")
+        logger.debug("Connected to mavsdk_server!")
 
     @property
     def channel(self):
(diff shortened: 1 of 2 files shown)


julianoes (commented):
Nice, yes that would be a great change. Do you think the default should be not to log these?

woosal1337 (approved):

qthibeault:
I think that the default should be to log the messages, since it is easier for users to discover and silence unwanted log messages rather than discover loggers to enable.

After looking at some other libraries like [urllib3](https://github.com/urllib3/urllib3) it looks like the convention is to not set a log level in the library modules and instead let users select the level when configuring logging for their application.

```python
logging.basicConfig(stream=stdout, level=logging.DEBUG)
```

julianoes:
@qthibeault ok so could we do it similar to urllib3 then?

qthibeault:
@julianoes I think that would be the best solution. I will remove the statement to set the level.

JonasVautherin (commented):
Sounds great! I just need to test it :blush:.

Out of curiosity, would you see an easy way to get the `mavsdk_server` output into that logger? Right now we just ignore it (see [here](https://github.com/mavlink/MAVSDK-Python/blob/main/mavsdk/system.py#L282-L285), but it would be great to have the choice.

qthibeault:
I added in functionality to log the output of the `mavsdk_server` process by configuring the sub-process to use a PIPE as `stdout` and creating a thread to monitor the pipe. 

Initially I tried to create a file-like object to use as `stdout`, but it appears that when given a file the subprocess library provides the file descriptor number not the file object to the subprocess so using virtual files doesn't work.

JonasVautherin (approved):
Working great for me, e.g. using like this:

```
diff --git a/examples/takeoff_and_land.py b/examples/takeoff_and_land.py
index 4eda81d..8ec98ce 100755
--- a/examples/takeoff_and_land.py
+++ b/examples/takeoff_and_land.py
@@ -1,6 +1,8 @@
 #!/usr/bin/env python3
 
 import asyncio
+import logging
+import sys
 from mavsdk import System
 
 
@@ -33,5 +35,7 @@ async def run():
     await drone.action.land()
 
 if __name__ == "__main__":
+    logging.basicConfig(stream=sys.stdout, level=logging.DEBUG)
+
     loop = asyncio.get_event_loop()
     loop.run_until_complete(run())
```

PR #391, opened 2021-09-16
macOS support for run_protoc.sh

mattes-bru:
I stumbled over an incompatibility with the macOS version of sed. I got this solved by doing this little hack using the homebrew gnu-sed package. There might be better solutions but I am neither a bash nor an regex expert but I think I'd better come with a solution than a problem. 😄

### DIFF
--- a/other/tools/run_protoc.sh
+++ b/other/tools/run_protoc.sh
@@ -24,7 +24,11 @@ command -v protoc-gen-mavsdk > /dev/null || {
 }
 
 function snake_case_to_camel_case {
-    echo $1 | sed -r 's/(^|_)([a-z])/\U\2/g'
+    if [[ "$OSTYPE" == "darwin"* ]]; then
+        echo $1 | gsed -r 's/(^|_)([a-z])/\U\2/g'
+    else
+        echo $1 | sed -r 's/(^|_)([a-z])/\U\2/g'
+    fi
 }
 
 function generate {
--- a/other/tools/run_protoc.sh
+++ b/other/tools/run_protoc.sh
@@ -24,11 +24,7 @@ command -v protoc-gen-mavsdk > /dev/null || {
 }
 
 function snake_case_to_camel_case {
-    if [[ "$OSTYPE" == "darwin"* ]]; then
-        echo $1 | gsed -r 's/(^|_)([a-z])/\U\2/g'
-    else
-        echo $1 | sed -r 's/(^|_)([a-z])/\U\2/g'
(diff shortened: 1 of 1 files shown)


other/tools/run_protoc.sh
```diff
@@ -24,7 +24,11 @@ command -v protoc-gen-mavsdk > /dev/null || {
 }
 
 function snake_case_to_camel_case {
-    echo $1 | sed -r 's/(^|_)([a-z])/\U\2/g'
+    if [[ "$OSTYPE" == "darwin"* ]]; then
```
julianoes, line 27: Where does `$OSTYPE` come from?

other/tools/run_protoc.sh
```diff
@@ -24,7 +24,11 @@ command -v protoc-gen-mavsdk > /dev/null || {
...
+        echo $1 | gsed -r 's/(^|_)([a-z])/\U\2/g'
```
julianoes, line 28: Does `gsed` exist on every Mac?

julianoes (commented):
Sorry, I was meaning to cherry-pick my fix that I did here:
https://github.com/mavlink/MAVSDK/commit/f158136941c3b92843058fdf780a94179bd6405c

Now we just need to decide which one we prefer.

other/tools/run_protoc.sh
mattes-bru: It is from one of the first search hits on "how to determine the os type in bash scripts" 😀 

It seems to be defined by default.

other/tools/run_protoc.sh
mattes-bru: No, it needs to be installed bye 'brew install gnu-sed'

mattes-bru:
Based on your comments and my responses, I would prefer yours, because it seems to be OS independent without any additional packages.

julianoes:
> Based on your comments and my responses, I would prefer yours, because it seems to be OS independent without any additional packages.

Me too. I was just trying to cherry-pick it but now realized it is already applied, however, that's in the C++ repo, not the Python one! Would you mind making a PR and using the awk magic?

mattes-bru:
@julianoes I pushed your suggestion to my branch. This should be enough to update this PR, otherwise let me know. I also tested it on my machine and it worked as expected.

julianoes (approved):

julianoes:
Great, thank you!

JonasVautherin (approved):

PR #397, opened 2021-10-21
renamed the equals function to make structs compareable by == operator

mattes-bru:
I stumbled over this issue that I can't compare GPSInfo. This should fix this.


JonasVautherin (approved):
Oh right, seems like __equals__ is not a thing. Thanks!

PR #407, opened 2021-11-30
system: allow setting sysid/compid

julianoes:
This allows to set the sysid and compid of the MAVSDK server but not actually the system, so the API is slightly odd.


JonasVautherin (approved):
Also it may be a bit confusing if you use another `mavsdk_server` than the embedded one, because those arguments won't have any effect then.

But well... :laughing:

PR #409, opened 2021-12-02
Add mavshell example

JonasVautherin:
To test (against a real board, SITL won't work):

1. Create a venv (or don't, but that's what I do):
    ```
    mkdir /tmp/test_mavshell && cd /tmp/test_mavshell
    python -m venv venv
    source ./venv/bin/activate
    ```
2. Install mavsdk (if not in a venv, use `pip install --user mavsdk` instead):
    ```
    pip install mavsdk
    ```
3. Run the example:
    ```
    python shell.py
    ```

Note: This assumes that you receive MAVLink on udp port 14540. If not, change the `await drone.connect(system_address="udp://:14540")` to listen to whatever you have (maybe `serial:///dev/ttyACM0`, maybe `udp://:14550`, you name it)

### DIFF
--- a/examples/shell.py
+++ b/examples/shell.py
@@ -1,0 +1,41 @@
+#!/usr/bin/env python3
+
+import asyncio
+import sys
+
+from mavsdk import System
+
+
+async def run():
+    drone = System()
+    await drone.connect(system_address="udp://:14540")
+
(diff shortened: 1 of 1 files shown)


julianoes:
You don't have in the instructions to copy the shell.py example there, right

julianoes (approved):
Nice!

Issue #369, opened 2021-07-27
Change home position (RTL position)


schpatz:
I want to change the home position (RTL position) so that it is different from the starting position. I thought about just adding it as the last mission item and then execute the `action.land()` command but that would have some complications with other commands that still use the RTL position.
Next I looked at the TelemetryServer Class and tried this:
```
telserver = telemetry_server.TelemetryServer(async_plugin_manager.AsyncPluginManager())
await telserver.publish_home(home=telemetry_server.Position(lat_home, lon_home, 35,0))
```
But then i get this error (probably because I'm not initializing it correctly):
`AttributeError: 'AsyncPluginManager' object has no attribute '_channel'`

Would be really greatful for some insight or a better way to achieve this :)

julianoes:
Ok there are at least two ways how you can force PX4 to land somewhere else than home.

1. You set the home position to a new position using the command [MAV_CMD_DO_SET_HOME](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_SET_HOME). This is currently not implemented in MAVSDK but would not be too hard. I'd think we could add it to the action plugin.
2. You use rally points which are explicitly made for the use case where you want "home positions" close to your drone. MAVSDK does not support rally points yet, however, the addition would not be too hard as the mechanism to add them is the same as adding mission items or geofence items. When using rally points, you can specify the behavior using the param [RTL_TYPE](https://docs.px4.io/master/en/advanced_config/parameter_reference.html#RTL_TYPE).

Let me know which approach you would prefer and then I can try to guide you in how to add the functionality.

schpatz:
> 1. You set the home position to a new position using the command [MAV_CMD_DO_SET_HOME](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_SET_HOME). This is currently not implemented in MAVSDK but would not be too hard. I'd think we could add it to the action plugin.

Thanks for the fast reply, that option sounds exactly like what I was looking for. I thought about that but didn't know how to send mavlink commands. I would be really greatful if you would walk me through how to implement the feature.

schpatz:
This seems really daunting and I will try to implement it, but it would be nice nevertheless if you or someone else tackles the problem as I am sure that some other people would benefit from this feature aswell :)

JonasVautherin:
Just a quick question: are we sure that MAV_CMD_DO_SET_HOME is implemented in PX4? Haven't checked, but I thought it was not

potaito:
[...]

PR #456, opened 2022-03-21
Same connection and GPS checks in all examples

potaito:
Context: https://github.com/mavlink/MAVSDK-Python/issues/369#issuecomment-1073729797

A year ago I extended the check prior to arming from 
```
if health.is_global_position_ok
```
to 
```
if health.is_global_position_ok and health.is_home_position_ok:
```
because otherwise the PX4 preflight checks might fail if the python scripts are too quickly attempting to arm the drone after it acquired GPS.

Unfortunately I added this patch to only a few examples and not all of them. This PR is making up for that, and it also unifies the connection check at the beginning of almost every script. I don't mind the exact phrasing of the messages and whether it's "drone discovered" or "drone connected", but I think it's good to have it similar in all example scripts.

### DIFF
--- a/examples/failure_injection.py
+++ b/examples/failure_injection.py
@@ -12,13 +12,13 @@ async def run():
     print("Waiting for drone to connect...")
     async for state in drone.core.connection_state():
         if state.is_connected:
-            print(f"Drone discovered with UUID: {state.uuid}")
+            print(f"-- Connected to drone!")
             break
 
     print("Waiting for drone to have a global position estimate...")
     async for health in drone.telemetry.health():
         if health.is_global_position_ok and health.is_home_position_ok:
-            print("Global position estimate ok")
+            print("-- Global position estimate OK")
             break
 
     print("-- Enabling failure injection")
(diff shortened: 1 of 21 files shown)


potaito:
@JonasVautherin please take a look :)

examples/failure_injection.py
```diff
@@ -12,13 +12,13 @@ async def run():
     print("Waiting for drone to connect...")
     async for state in drone.core.connection_state():
         if state.is_connected:
-            print(f"Drone discovered with UUID: {state.uuid}")
+            print(f"-- Connected to drone!")
```
potaito, line 15: If I recall correctly, `{state.uuid}` was causing issues with some MAVSDK versions. At least I got errors sometimes trying to run this script. Should we just remove it or is there a new "proper way" to access the UUID?

JonasVautherin: Now the UUID is part of the info plugin, I think :+1:

JonasVautherin (approved):
Thanks!

PR #461, opened 2022-04-01
[WIP] Add status text example

danylevskyi:
There are no examples for sending status using the `server_utility` plugin.

The work in progress.
The code is being executed, but QGroundControll doesn't receive any message.

### DIFF
--- a/examples/takeoff_and_land.py
+++ b/examples/takeoff_and_land.py
@@ -2,6 +2,7 @@
 
 import asyncio
 from mavsdk import System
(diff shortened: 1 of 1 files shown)


JonasVautherin:
Nice, thanks for the contribution!

Could we consider putting that in another example? I'd like to keep `takeoff_and_land` very small, as it's supposed to be a simple "hello world" :blush:.

danylevskyi:
Just added new example: https://github.com/mavlink/MAVSDK-Python/pull/465

PR #465, opened 2022-04-02
Add example for server_utility plugin

danylevskyi:
image


julianoes (approved):
Very nice, thank you!

PR #471, opened 2022-04-26
Pr fix windows installation

lyraik:
Fixed the executables extention to start the mavlink server on windows
@JonasVautherin

### DIFF
--- a/mavsdk/system.py
+++ b/mavsdk/system.py
@@ -352,7 +352,12 @@ class System:
             from importlib_resources import path
 
         try:
-            with path(bin, 'mavsdk_server') as backend:
+            if sys.platform.startswith('win'):
+                mavsdk_exec_name = "mavsdk_server.exe"
+            else:
+                mavsdk_exec_name = "mavsdk_server"
+
+            with path(bin, mavsdk_exec_name) as backend:
                 bin_path_and_args = [os.fspath(backend),
                                      "-p", str(port),
                                      "--sysid", str(sysid),
--- a/setup.py
+++ b/setup.py
@@ -76,8 +76,12 @@ class custom_build(build):
     def mavsdk_server_filepath(self):
         """
         The location of the downloaded `mavsdk_server` binary
+        For Windows this needs to be a .exe file
         """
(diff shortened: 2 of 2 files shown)


setup.py
```diff
@@ -76,9 +76,14 @@ def platform_suffix(self):
     def mavsdk_server_filepath(self):
         """
         The location of the downloaded `mavsdk_server` binary
-        """
+        For windows this needs to be a .exe file
+        """   
+        if sys.platform.startswith('win'):
+            return 'mavsdk/bin/mavsdk_server.exe'
         return 'mavsdk/bin/mavsdk_server'
```
JonasVautherin, line 83: Nitpick: I think this is slightly more readable:

```suggestion
        if sys.platform.startswith('win'):
            return 'mavsdk/bin/mavsdk_server.exe'
        else:
            return 'mavsdk/bin/mavsdk_server'
```

mavsdk/system.py
```diff
@@ -352,21 +352,24 @@ def _start_mavsdk_server(system_address, port, sysid, compid):
...
+            if system_address:
+                bin_path_and_args.append(system_address)
+            p = subprocess.Popen(bin_path_and_args,
+                                    shell=False,
+                                    stdout=subprocess.PIPE,
+                                    stderr=subprocess.STDOUT)
+
+            logger = logging.getLogger(__name__)
+            log_thread = _LoggingThread(p.stdout, logger.debug)
+            log_thread.start()
```
JonasVautherin, line 372: hmm are you entirely sure that changing the indentation here does not change the behavior?

I'd rather keep the old indentation that I know works, if there is not a good reason to change it :blush:

JonasVautherin (changes requested):
Thanks a lot for the contribution! Two small remarks, though :innocent:

setup.py
lyraik: I had It like your change before and it was suggested I change it to the other one 😆

JonasVautherin (approved):
Thanks a lot! :rocket:

lyraik:
Thanks for adding the changes @JonasVautherin

PR #472, opened 2022-04-28
setup.py: drop setup_requires

dlech:
The requirements in `requirements-dev.txt` are not actually required to run `setup.py` so this can be removed.


JonasVautherin (approved):
Thanks!

PR #477, opened 2022-05-27
Delete follow2.py

potaito:
Empty file :)


JonasVautherin (changes requested):
Hmm not sure about this, it feels very risky. Let's keep it for now.

JonasVautherin (approved):
Just kidding :grin:. Thanks @potaito!

PR #518, opened 2022-09-30
Fixed ComponentInformationServer key error

carlsondev:
The `component_information_server` property in `system.py` uses the wrong key when returning the value leading to a KeyError. This changes that key to the proper one.


JonasVautherin (approved):
Thanks :)

PR #528, opened 2022-11-01
example: add status_text example

julianoes:
Works for me.

@JonasVautherin note this has the new way of running stuff with `asyncio` in Python 3.10.


JonasVautherin (approved):

JonasVautherin:
> note this has the new way of running stuff with asyncio in Python 3.10.

Oh they added a simplified `asyncio.run` instead of having to get the event loop? Nice :+1:

PR #536, opened 2022-12-21
readme: clarify on editable installation

potaito:
It went over my head that `-e` does not copy files to the system's default location until I googled it. Should we add this extra bit of information to the README?


JonasVautherin (approved):
Thanks!

PR #538, opened 2022-12-22
Fix macos ci

JonasVautherin:
For some reason it started failing with `error: invalid command 'bdist_wheel'`.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -125,7 +125,7 @@ jobs:
 
   macOS:
     name: macOS
-    runs-on: macos-latest
+    runs-on: macos-12
     steps:
       - uses: actions/checkout@v1
         with:
@@ -140,7 +140,7 @@ jobs:
       - name: Install prerequisites
         run: |
           python3 -m pip install -r requirements.txt
-          python3 -m pip install twine
+          python3 -m pip install twine wheel
           python3 -m pip install --user delocate
 
       - name: Create wheel


JonasVautherin:
I am a bit puzzled, because it was very painful to get to those two lines. I am not entirely sure if those two lines generally work or if I have just been lucky in the last 4 runs.

At least they make sense, so... let's try with them maybe?

.github/workflows/main.yml
```diff
@@ -125,7 +125,7 @@ jobs:
 
   macOS:
     name: macOS
-    runs-on: macos-latest
+    runs-on: macos-12
```
julianoes, line 128: So according to what I read this should be the same. Although I'm fine pinning this to avoid breakage.

.github/workflows/main.yml
```diff
@@ -140,7 +140,7 @@ jobs:
       - name: Install prerequisites
         run: |
           python3 -m pip install -r requirements.txt
-          python3 -m pip install twine
+          python3 -m pip install twine wheel
```
julianoes, line 143: I've seen that before, probably makes sense.

Why did you remove `twine`? I believe that's required to publish the release, no?

.github/workflows/main.yml
JonasVautherin: Yep, but before I changed it, it was failing :thinking:. So I don't know if something else changed during my tests, or if that is different, but I'm a bit confused.

.github/workflows/main.yml
JonasVautherin: I did not, it's right before `wheel` :laughing:. What time was it when you read that? :grin:

julianoes: Gosh. Thanks.

PR #551, opened 2023-01-30
examples: add simple mission_raw example

julianoes:
As an example how to set seq and current.


JonasVautherin (approved):
Awesome! :rocket:

PR #565, opened 2023-03-26
Fix firmware_version.py file

arpitpara:
the request adds changes with respect to recent update in mavsdk_error for version infor error.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-    "ros.distro": "foxy"
-}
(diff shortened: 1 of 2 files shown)


JonasVautherin:
Thanks for the contribution!

Can you elaborate a bit on it? What changed in MAVSDK? Your PR suggests that the first call somehow always fails, but that should not be the case :thinking:.

arpitpara:
> Thanks for the contribution!
> 
> Can you elaborate a bit on it? What changed in MAVSDK? Your PR suggests that the first call somehow always fails, but that should not be the case thinking.

Sorry for minor misatke!
I forgot to add the link of issue fix and await statement before getting info.

julianoes:
This is not needed once we update mavsdk_server to v1.4.13 which I plan to do ASAP.

Issue #556, opened 2023-02-10
Unable to connect Pixhawk to Raspberry Pi using MAVSDK 


aadithyasujit:
Hi @julianoes, 
I am trying to connect my Pixhawk 4 (running PX4) to Raspberry Pi 3/4 using MAVSDK but end up with the below error. I have installed MAVSDK using the pip3 command following instructions from [here](http://mavsdk-python-docs.s3-website.eu-central-1.amazonaws.com/index.html#install-using-pip-from-pypi). 
![image](https://user-images.githubusercontent.com/33419672/218153101-e401f47b-9157-4a3c-af1b-96a582bf7ba0.png)
I am using an FTDI adapter and connecting the pixhawk to Pi using usb & use the below command to connect to the Pixhawk
![image](https://user-images.githubusercontent.com/33419672/218154241-95710bf4-6a61-4891-b2ed-1e9665146056.png)

In an attempt to troubleshoot, I am able to receive data on the port when i checked screen using screen /dev/ttyUSB0 and when I try the connecting using using Mavproxy, it works fine. But I have not been able to establish a connection MAVSDK. 
What else would I have to check in order to get this working with MAVSDK?

julianoes:
You might need the latest bugfix: https://github.com/mavlink/MAVSDK/pull/1968
but it's not released in Python yet.

Let me do a Python release and then you can try again.

julianoes:
Can you try with https://github.com/mavlink/MAVSDK-Python/releases/tag/1.4.4. It should be on PyPi shortly.

aadithyasujit:
> Can you try with https://github.com/mavlink/MAVSDK-Python/releases/tag/1.4.4. It should be on PyPi shortly.

I did try this using `pip3 install mavsdk --upgrade`. 
```
Collecting mavsdk
  Downloading mavsdk-1.4.4-py3-none-manylinux_2_5_x86_64.manylinux1_x86_64.manylinux_2_17_x86_64.manylinux2014_x86_64.whl (11.6 MB)
     |████████████████████████████████| 11.6 MB 1.6 MB/s 
Requirement already satisfied, skipping upgrade: grpcio>=1.50.0 in /home/linux-sim/.local/lib/python3.8/site-packages (from mavsdk) (1.51.1)
Requirement already satisfied, skipping upgrade: aiogrpc>=1.8 in /home/linux-sim/.local/lib/python3.8/site-packages (from mavsdk) (1.8)
[...]

PR #566, opened 2023-03-26
firware_version.py update

arpitpara:
The request adds on the required function according to recent update of mavsdk_server 1.4.13.
I am adding the respective issue link for the reference.
[Issue link](https://github.com/mavlink/MAVSDK-Python/issues/556)
Mavsdk fix - https://github.com/mavlink/MAVSDK/pull/2006

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "ros.distro": "foxy"
+}
--- a/examples/firmware_version.py
+++ b/examples/firmware_version.py
@@ -17,6 +17,7 @@ async def run():
     try:
         info = await drone.info.get_version()
     except:
+        await asyncio.sleep(2)
         info = await drone.info.get_version()
         print(info)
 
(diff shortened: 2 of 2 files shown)


JonasVautherin:
My understanding was that this was fixed in MAVSDK, so that it does not need to be changed in the examples :thinking:. Am I missing something?

arpitpara:
> My understanding was that this was fixed in MAVSDK, so that it does not need to be changed in the examples thinking. Am I missing something?
> 
> I mean, what's the point of waiting 1.5s in MAVSDK if you then wait 2s in the example?

Actually, even I thought so
Let me check by updating the masdk version.

arpitpara:
mavsdk version is still 1.4.13 in MAVSDK Python

arpitpara:
Yeah It is working with newer version.
Sorry for bothering. Just started in this stuff so thanks for the remarks.

arpitpara:
Thanks

julianoes:
Fixed in 1.4.5.

Issue #585, opened 2023-04-24
mavsdk_server errors not shown to users


EricPedley:
I was using mavsdk to try communicating with PX4 SITL today and also had mavros running, which made mavsdk_server crash on launch because the udp connection I was trying was already in use. It was hard to debug because when `system.py` runs `mavsdk_server` as a subprocess, the stdout is redirected to a logging thread which logs everything at debug level, which is not shown to users (and for new users like me I don't even know how to get it to show me the debug output). From the user's perspective, the call to `connect` just hangs forever. This error also happens if you input a malformed URL like `udp://14540` (which is missing a colon before the port number), it'll just hang forever instead of giving the user a useful error message.

JonasVautherin:
Wasn't it already addressed by [this PR](https://github.com/mavlink/MAVSDK-Python/pull/370)? Would you mind trying it and letting me know if that solves your problem?

I had tried it with an example script [here](https://github.com/mavlink/MAVSDK-Python/pull/370#pullrequestreview-729712188).

EricPedley:
Tried that and for some reason it still doesn't log the error messages. I'm running this program to test out the logging:
```
import asyncio
from mavsdk import System
import logging
import sys

logging.basicConfig(stream=sys.stdout, level=logging.DEBUG)

async def main():
    drone = System()
    print("awaiting connection")
    await drone.connect(system_address="udp://14550") # should give an error message about bad connection URL
    print("done connecting")
    version_info = await drone.info.get_version()
    print(version_info)

asyncio.run(main())
```
The output with mavsdk-python on main looks like this:
![image](https://user-images.githubusercontent.com/48658337/234181435-cf8f2c5d-29d1-49b1-92d6-fd973d6a2588.png)
And with the changes in my PR (#586) it looks like this:
![image](https://user-images.githubusercontent.com/48658337/234181513-af5dc80a-83e9-4fbf-b67b-b7760af14089.png)
Either way, I feel like #370 isn't sufficient here because the output from `mavsdk_server` seems important enough to always log by default, but for it to be shown the user needs to turn on debug-level logging.

JonasVautherin:
> I think that something is broken with the logging thread for mavsdk_server because it should be displaying the stuff when I turn on debug logging

Yes it should, I remember it was working for me :thinking:. Maybe it got broken somehow. Would you be able to look into it? I'll try to reproduce it when I have time :+1:.

[...]

PR #586, opened 2023-04-24
Simplify logging of mavsdk_server output

EricPedley:
# Motivation
In #585 I described how logging output from the child process for `mavsdk_server` at only debug level hides useful errors from the user.

# Summary
This PR removes the logic for redirecting the subprocess stdout to the debugger and results in the mavsdk_server child process's stdout and stderr going straight to the main stdout/stderr.

### DIFF
--- a/mavsdk/system.py
+++ b/mavsdk/system.py
@@ -74,7 +74,7 @@ class System:
     def __del__(self):
         self._stop_mavsdk_server()
 
-    async def connect(self, system_address=None):
+    async def connect(self, system_address=None, verbose=True):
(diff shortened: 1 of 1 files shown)


julianoes (commented):
What about making it an option?

EricPedley:
> What about making it an option?

Yeah this is a good idea. I added another argument to `System.connect` that lets users turn off this logging.

JonasVautherin:
I think we should try to fix what was done in #370 (if it is indeed broken) instead :thinking:

PR #602, opened 2023-06-05
examples: add RTK example

julianoes:
This shows how RTCM corrections can be sent using the Rtk plugin.

This will only work starting with MAVSDK server version https://github.com/mavlink/MAVSDK/releases/tag/v1.4.16 and thus MAVSDK-Python v1.4.8.

### DIFF
--- a/examples/rtk.py
+++ b/examples/rtk.py
@@ -1,0 +1,55 @@
+#!/usr/bin/env python3
+
+import asyncio
+from mavsdk import System, rtk
+
+# Some sample data that won't work because it needs to have the correct time
+# information. However, that's the sort of bytes that can be expected.
+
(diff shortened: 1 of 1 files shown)


dayjaby:
Looking good! Maybe we could extend this with some code reading the RTCM data from an actual GPS device, like a ublox chip? Or would that be too specific?

JonasVautherin (approved):

julianoes:
@dayjaby yes please, if you could add something like that, that would be fantastic. Or maybe there is some pip package that helps with connecting to ntrip.

dayjaby:
@julianoes sure, but feel free to bring your example already in. I will create another PR for that.

PR #604, opened 2023-06-10
examples: add RTK example with uBlox base

dayjaby:
This example has a few prerequisites:

1. pyserial is installed
2. the uBlox base is connected via USB (otherwise the correct serial device has to be used in the code, or should we make it a command line argument?)
3. the uBlox base is properly configured. This code works even if it receives UBX/NMEA messages, but the user itself is in charge to configure the RTCM messages to be sent out. 
4. this code can directly connect to the drone; which means it cannot fly like this if attached via USB to the RTK base: Do we make the assumption everyone knows how to route mavlink?

Regarding 3: We could add some code to configure the uBlox device automatically via pyubx2 for example.

Tested against PX4 simulation:
image

### DIFF
--- a/examples/rtk_base_ublox.py
+++ b/examples/rtk_base_ublox.py
@@ -82,7 +82,15 @@ class RTCMParser:
 
 
 async def run():
-    # Init the drone
+    # Init the drone.
+    # Note that connecting directly to a drone is not very useful here.
+    # You probably want to set-up some MAVLink message routing inbetween
+    # this script and the drone.
+    # Good candidates for this are
+    # 1. mavlink-router (https://github.com/mavlink-router/mavlink-router)
+    # 2. MAVSDK (https://tinyurl.com/mavsdk-forwarding)
+    # In the latter case you can use something like this here:
+    # await drone.connect("udp://<IP address of the drone here>:14540")
     drone = System()
     await drone.connect()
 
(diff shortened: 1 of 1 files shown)


julianoes:
Great, thanks @dayjaby! I would suggest to add your notes to the script so that readers see it.

examples/rtk_base_ublox.py
```diff
@@ -0,0 +1,148 @@
...
+
+async def print_gps_info(drone):
+    async for gps_info in drone.telemetry.gps_info():
+        print(f"GPS info: {gps_info}")
+
+
+async def send_rtcm(drone):
+
+    # Connect to a ublox RTK base station via USB
+    ublox = serial.Serial("/dev/ttyACM0", 9600)
```
julianoes, line 105: Wouldn't the baudrate usually be higher?

dayjaby: iirc that's the default and enough for RTCM: 3 messages at 1Hz, and if survey completed 1 message at 0.1Hz. It still has room for plenty UBX messages

dayjaby:
> Great, thanks @dayjaby! I would suggest to add your notes to the script so that readers see it.

I tried to cover them in the latest commit.

Just an idea for routing: We could add another executable to the .debs in MAVSDK, which does nothing but routing:

```mavsdk_router udp://0.0.0.0:14540 serial:///dev/ttyACM0```

This would help the pure python MAVSDK users with the routing issue and it's super small (~27kB here).

julianoes (approved):
Thanks!

julianoes:
> We could add another executable to the .debs in MAVSDK, which does nothing but routing:

I would argue mavlink-router could provide that. I'd rather not have that as a dependency.

PR #609, opened 2023-06-21
Example-EV/GNSS fusion force switching

farhangnaderi:
@afwilkin has proposed that we can introduce a way with MAVSDK-Python example script to force switch between 3D-Pose fusion sources (VIO/GPS) for the special cases which these scenarios are needed. This also could be useful when users need to test their EV fusion and it healthy state. 
We are aware that EKF2 does this switching automatically but based on the above reasons we came to the point that it could also be useful to have specific examples to swap between fusion means. 
I have come to the point of two different kinds of examples, one with Keyboard input and another one with time-based method. 

@julianoes I would appreciate if you review this and share your ideas.

### DIFF
--- a/examples/ev_automation.py
+++ b/examples/ev_automation.py
@@ -1,0 +1,68 @@
+import asyncio
+from mavsdk import System
+
+async def set_params(system, params, announcement):
+    print(announcement)
+
+    async for _ in system.telemetry.landed_state():
+        break
+
+    for param, value in params:
+        await system.param.set_param_int(param, value)
+        await asyncio.sleep(5)
+
+    async for state in system.telemetry.landed_state():
+        break
+
+    print(f"Landed state: {state}")
+
+async def main():
+    drone = System()
+    await drone.connect(system_address="udp://:14540")
+
+    print("Waiting for drone to connect...")
+    async for state in drone.core.connection_state():
+        if state.is_connected:
+            print("-- Connected to drone!")
+            break
+
+    params_preflight = [
+        ("EKF2_HGT_REF", 0),
+        ("EKF2_EV_CTRL", 15),
+        ("EKF2_GPS_CTRL", 7)
+    ]
+    await set_params(drone, params_preflight, "Setting preflight parameters...")
+
+    # Wait for 5 seconds
+    await asyncio.sleep(5)
+
+    params_gps_required = [
+        ("EKF2_HGT_REF", 1),
+        ("EKF2_EV_CTRL", 0),
+        ("EKF2_GPS_CTRL", 7)
+    ]
+    await set_params(drone, params_gps_required, "Setting airborne (GPS Required) parameters...")
+
+    # Wait for 10 seconds
+    await asyncio.sleep(10)
+
+    params_ev_required = [
+        ("EKF2_HGT_REF", 3),
+        ("EKF2_EV_CTRL", 15),
+        ("EKF2_GPS_CTRL", 0)
+    ]
+    await set_params(drone, params_ev_required, "Setting airborne (EV Required) parameters...")
+
+    # Start the tasks to print flight mode and system status
+    print_mode_task = asyncio.ensure_future(print_mode(drone))
+    print_status_task = asyncio.ensure_future(print_status(drone))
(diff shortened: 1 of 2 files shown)


examples/ev_automation.py
```diff
@@ -0,0 +1,68 @@
+import asyncio
```
julianoes, line 1:
```suggestion
#!/usr/bin/env python

import asyncio
```

examples/ev_automation_keyboard.py
```diff
@@ -0,0 +1,74 @@
+import asyncio
```
julianoes, line 1:
```suggestion
#!/usr/bin/env python

import asyncio
```

examples/ev_automation_keyboard.py
```diff
@@ -0,0 +1,74 @@
...
+
+async def set_params(system, params, announcement):
+    print(announcement)
+
+    async for _ in system.telemetry.landed_state():
+        break
+
+    for param, value in params:
+        await system.param.set_param_int(param, value)
+        await asyncio.sleep(5)
```
julianoes, line 12: Why this sleep? That shouldn't be required, right?

julianoes (changes requested):
Thanks for the examples @farhangnaderi.

It would be good if you could rename them to start with `px4_` because given the params these are PX4 specific.

Also, make sure the style check passes, and there is a small comment at the top of the file explaining what it does.

examples/ev_automation_keyboard.py
farhangnaderi: Unnecessary, removed! Thanks for noticing it!

farhangnaderi:
We might be good to go now?

examples/px4_ev_automation.py
```diff
@@ -1,21 +1,21 @@
+#!/usr/bin/env pytho
```
julianoes, line 1:
```suggestion
#!/usr/bin/env python
```

examples/px4_ev_automation_keyboard.py
```diff
@@ -1,21 +1,21 @@
+#!/usr/bin/env pytho
```
julianoes, line 1:
```suggestion
#!/usr/bin/env python
```

julianoes (changes requested):

julianoes:
@farhangnaderi 

Please make sure there is a small comment at the top of the file explaining what it does. Otherwise, how would a reader know what this does? (I don't even know what EV stands for :wink:)

farhangnaderi:
> @farhangnaderi
> 
> Please make sure there is a small comment at the top of the file explaining what it does. Otherwise, how would a reader know what this does? (I don't even know what EV stands for 😉)

Thanks Julian! this happens when I am making a PR in drunk mode (but not necessarily being drunk) :)
Thanks for the review! Is the description clear? Shall I add anything more to it?

examples/px4_ev_automation.py
```diff
@@ -1,5 +1,16 @@
 #!/usr/bin/env python
 
+"""
+
+This example can be used to switch between Eaxtenal Vision or MOCAP (EV) fusion and GNSS data fusion in PX4 
+firmware (v1.14 and on) with a time-based switching.
```
julianoes, line 6:
```suggestion
This example can be used to switch between External Vision or MOCAP (EV) fusion and GNSS data fusion in PX4 
firmware (v1.14 and later) with time-based switching.
```

examples/px4_ev_automation_keyboard.py
```diff
@@ -0,0 +1,85 @@
+#!/usr/bin/env python
+
+"""
+
+This example can be used to switch between Eaxtenal Vision or MOCAP (EV) fusion and GNSS data fusion in PX4 
+firmware (v1.14 and on) by using kyeboard inputs from the user.
```
julianoes, line 6: Same here...

julianoes: And check CI.

farhangnaderi:
Closing due to [#611 ](https://github.com/mavlink/MAVSDK-Python/pull/611)

PR #619, opened 2023-09-13
Update manual_control.py

alanmehio:
Fix the bug first. Later we should fix the error which comes when executing it against  SITL ( jmavsim simulation )


alanmehio:
Fix the bug first .

JonasVautherin (approved):

julianoes (approved):

julianoes:
Thanks @alanmehio!

PR #669, opened 2024-03-19
Update mission.py

MuhammadBilal1:
A new parameter is added to the "MissionItem" constructor, but that parameter was mission from the example file which was causing the error to appear upon running the mission.py file. I added the missing parameter at the appropriate location.


julianoes (approved):

julianoes:
Thanks!

PR #673, opened 2024-03-24
Geofence mechanism is updated according to V2.1.0.

MuhammadBilal1:
The process of setting the geofence in V1 was different than V2, so this change updated the geofence process according to the need of V2. Geofence class is changed.


julianoes (approved):
Nice, thank you!

PR #727, opened 2024-09-03
Example to print IMU data

dr0ne-guy:
I have created a python file that prints the IMU data from the FC

### DIFF
--- a/examples/highres_imu.py
+++ b/examples/highres_imu.py
@@ -1,0 +1,43 @@
+import asyncio
+from mavsdk import System
+
+async def get_imu_data():
+    # Connect to the drone
+    drone = System()
+    await drone.connect(system_address="serial:///dev/ttyTHS0:3000000")
+
+    # Wait for the drone to connect
+    print("Waiting for drone to connect...")
+    async for state in drone.core.connection_state():
+        if state.is_connected:
+            print("Drone is connected!")
+            break
+
+    telemetry = drone.telemetry
+
+    # Set the rate at which IMU data is updated (in Hz)
+    await telemetry.set_rate_imu(200.0)
+
+    # Fetch and print IMU data
+    print("Fetching IMU data...")
+    async for imu in telemetry.imu():
+        # Print data in HIGHRES_IMU format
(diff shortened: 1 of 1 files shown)


examples/highres_imu.py
```diff
@@ -0,0 +1,43 @@
...
+        print(f"Z Acceleration (m/s^2): {imu.acceleration_frd.down_m_s2}")
+        print(f"X Gyro (rad/s): {imu.angular_velocity_frd.forward_rad_s}")
+        print(f"Y Gyro (rad/s): {imu.angular_velocity_frd.right_rad_s}")
+        print(f"Z Gyro (rad/s): {imu.angular_velocity_frd.down_rad_s}")
+        print(f"X Mag (gauss): {imu.magnetic_field_frd.forward_gauss}")
+        print(f"Y Mag (gauss): {imu.magnetic_field_frd.right_gauss}")
+        print(f"Z Mag (gauss): {imu.magnetic_field_frd.down_gauss}")
+        print(f"Temperature (°C): {imu.temperature_degc}")
+        print("-----------------------------------------")
+        await asyncio.sleep(0)  # Sleep for a second between each print
```
JonasVautherin, line 38: This is not sleeping for a second, and in any case you should not sleep here :see_no_evil:

JonasVautherin (changes requested):

dr0ne-guy:
@JonasVautherin I have changed the sleep timer.

JonasVautherin:
But you should not sleep there :see_no_evil:. You should consume the events as they arrive. Sleeping may just slow them down and accumulate a delay that you don't want :innocent:.

dr0ne-guy:
I have removed the delay 😅

julianoes (approved):
Thanks for that!

julianoes:
If you can fix the style issue then this is good to go in, thanks.
https://github.com/mavlink/MAVSDK-Python/actions/runs/10682857899/job/29632558957?pr=727

dr0ne-guy:
I have changed the style

PR #758, opened 2025-04-07
Register log_streaming plugin

rantalaine:
Add log_streaming as a property to System and initialize it along with rest of the plugins


julianoes (approved):
Thanks!

julianoes:
I think an example could help in implementing this.

Similar to:
https://github.com/mavlink/MAVSDK/blob/main/examples/log_streaming/log_streaming.cpp

PR #769, opened 2025-06-29
useless-semicolon: `ruff check --select=E703 --fix`

cclauss:
% `ruff check --select=E703 --fix`
```
Found 18 errors (18 fixed, 0 remaining).
```
% [`ruff rule E703`](https://docs.astral.sh/ruff/rules/useless-semicolon)
# useless-semicolon (E703)

Derived from the **pycodestyle** linter.

Fix is always available.

## What it does
Checks for statements that end with an unnecessary semicolon.

## Why is this bad?
A trailing semicolon is unnecessary and should be removed.

## Example
```python
do_four();  # useless semicolon
```

Use instead:
```python
do_four()
```


julianoes (approved):
Thanks!

PR #786, opened 2025-09-03
Ruff fixes

### LINKED ISSUE #771: Linting issues for asyncio
I have little expertise in using asyncio.  Are any of these problematic?

% `ruff check --select=ASYNC,RUF006 --statistics`
```
16	RUF006  	asyncio-dangling-task
 3	ASYNC110	async-busy-wait
 2	ASYNC230	blocking-open-call-in-async-function
~1	ASYNC251	blocking-sleep-in-async-function~  # Fixed in #772
Found 22 errors.
```
% `ruff rule RUF006 ` # https://docs.astral.sh/ruff/rules/asyncio-dangling-task
# asyncio-dangling-task (RUF006)

Derived from the **Ruff-specific rules** linter.

## What it does
Checks for `asyncio.create_task` and `asyncio.ensure_future` calls
that do not store a reference to the returned result.

## Why is this bad?
Per the `asyncio` documentation, the event loop only retains a weak
reference to tasks. If the task returned by `asyncio.create_task` and
`asyncio.ensure_future` is not stored in a variable, or a collection,
or otherwise referenced, it may be garbage collected at any time. This
can lead to unexpected and inconsistent behavior, as your tasks may or
may not run to completion.

## Example
```python
import asyncio

for i in range(10):
    # This creates a weak reference to the task, which may be garbage
    # collected at any time.
    asyncio.create_task(some_coro(param=i))
```

Use instead:
```python
import asyncio

background_tasks = set()

for i in range(10):
    task = asyncio.create_task(some_coro(param=i))

    # Add task to the set. This creates a strong reference.
    background_tasks.add(task)

    # To prevent keeping references to finished tasks forever,
    # make each task remove its own reference from the set after
    # completion:
    task.add_done_callback(background_tasks.discard)
```

## References
- [_The Heisenbug lurking in your async code_](https://textual.textualize.io/blog/2023/02/11/the-heisenbug-lurking-in-your-async-code/)
- [The Python Standard Library](https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task)

julianoes:
Closes #771.

@cclauss

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -26,9 +26,10 @@ jobs:
 
       - name: Check style
         run: |
-          python3 -m pip install pycodestyle
+          python3 -m pip install pycodestyle ruff
           export PATH=$PATH:$HOME/.local/bin
           pycodestyle examples/*
+          ruff --select=ASYNC,RUF006
 
       - name: Install prerequisites
         run: |
--- a/examples/camera_params.py
+++ b/examples/camera_params.py
@@ -24,9 +24,10 @@ async def run():
     drone = System()
     await drone.connect(system_address="udpin://0.0.0.0:14540")
 
-    asyncio.ensure_future(observe_current_settings(drone))
-    asyncio.ensure_future(observe_camera_mode(drone))
-    asyncio.ensure_future(observe_possible_setting_options(drone))
+    tasks = []
+    tasks.append(asyncio.create_task(observe_current_settings(drone)))
+    tasks.append(asyncio.create_task(observe_camera_mode(drone)))
+    tasks.append(asyncio.create_task(observe_possible_setting_options(drone)))
 
     while True:
         entered_input = await ainput(usage_str)
--- a/examples/offboard_position_velocity_ned.py
+++ b/examples/offboard_position_velocity_ned.py
@@ -42,7 +42,7 @@ async def run():
         async for odom in drone.telemetry.position_velocity_ned():
             print(f"{odom.velocity.north_m_s} {odom.velocity.down_m_s}")
 
-    asyncio.ensure_future(print_z_velocity(drone))
+    tasks = {asyncio.create_task(print_z_velocity(drone))}
 
     print("-- Go 0m North, 0m East, -10m Down within local coordinate system")
     await drone.offboard.set_position_velocity_ned(
--- a/examples/rtk_base_ublox.py
+++ b/examples/rtk_base_ublox.py
@@ -99,8 +99,9 @@ async def run():
     asyncio.ensure_future(print_gps_info(drone))
     asyncio.ensure_future(send_rtcm(drone))
 
-    while True:
-        await asyncio.sleep(1)
+    # Keep the program running indefinitely
+    exit_event = asyncio.Event()
+    await exit_event.wait()
 
 
 async def print_gps_info(drone):
--- a/examples/rtk_base_ublox.py
+++ b/examples/rtk_base_ublox.py
@@ -96,8 +96,9 @@ async def run():
     await drone.connect()
 
     # Start the tasks
-    asyncio.ensure_future(print_gps_info(drone))
-    asyncio.ensure_future(send_rtcm(drone))
+    tasks = []
+    tasks.append(asyncio.create_task(print_gps_info(drone)))
+    tasks.append(asyncio.create_task(send_rtcm(drone)))
 
     # Keep the program running indefinitely
     exit_event = asyncio.Event()
--- a/examples/shell.py
+++ b/examples/shell.py
@@ -5,12 +5,14 @@ import sys
 
 from mavsdk import System
 
+send_tasks = set()
+
 
 async def run():
     drone = System()
     await drone.connect(system_address="udpin://0.0.0.0:14540")
 
-    asyncio.ensure_future(observe_shell(drone))
+    tasks = {asyncio.create_task(observe_shell(drone))}
 
     print("Waiting for drone to connect...")
     async for state in drone.core.connection_state():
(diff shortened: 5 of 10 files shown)


.github/workflows/main.yml
```diff
@@ -26,9 +26,10 @@ jobs:
 
       - name: Check style
         run: |
-          python3 -m pip install pycodestyle
+          python3 -m pip install pycodestyle ruff
           export PATH=$PATH:$HOME/.local/bin
           pycodestyle examples/*
+          ruff --select=ASYNC,RUF006
```
cclauss, line 32:
```suggestion
          ruff check --select=ASYNC,RUF006
```

julianoes: Thanks!

examples/camera_params.py
```diff
@@ -24,7 +24,7 @@ async def run():
     drone = System()
     await drone.connect(system_address="udpin://0.0.0.0:14540")
 
-    tasks = [
+    _tasks = [
```
cclauss, line 27: It is unusual to assemble these tasks but not await them.  Are we certain that they will be run at some point?

examples/offboard_position_velocity_ned.py
```diff
@@ -42,7 +42,7 @@ async def print_z_velocity(drone):
         async for odom in drone.telemetry.position_velocity_ned():
             print(f"{odom.velocity.north_m_s} {odom.velocity.down_m_s}")
 
-    tasks = {asyncio.create_task(print_z_velocity(drone))}
+    _tasks = {asyncio.create_task(print_z_velocity(drone))}
```
cclauss, line 45: It is unusual to assemble these tasks but not await them.  Are we certain that they will be run at some point?

examples/rtk_base_ublox.py
```diff
@@ -96,7 +96,7 @@ async def run():
     await drone.connect()
 
     # Start the tasks
-    tasks = [
+    _tasks = [
```
cclauss, line 99: It is unusual to assemble these tasks but not await them.  Are we certain that they will be run at some point?

examples/shell.py
```diff
@@ -12,12 +12,12 @@ async def run():
     drone = System()
     await drone.connect(system_address="udpin://0.0.0.0:14540")
 
-    tasks = {asyncio.create_task(observe_shell(drone))}
+    _tasks = {asyncio.create_task(observe_shell(drone))}
```
cclauss, line 15: It is unusual to assemble these tasks but not await them.  Are we certain that they will be run at some point?

examples/telemetry.py
```diff
@@ -15,7 +15,7 @@ async def run():
     await drone.connect(system_address="udpin://0.0.0.0:14540")
 
     # Start the tasks
-    tasks = [
+    _tasks = [
```
cclauss, line 18: It is unusual to assemble these tasks but not await them.  Are we certain that they will be run at some point?

examples/transponder.py
```diff
@@ -10,7 +10,7 @@ async def run():
     await drone.connect()
 
     # Start the tasks
-    tasks = {asyncio.create_task(print_transponders(drone))}
+    _tasks = {asyncio.create_task(print_transponders(drone))}
```
cclauss, line 13: It is unusual to assemble these tasks but not await them.  Are we certain that they will be run at some point?

.github/workflows/main.yml
```diff
@@ -27,7 +27,7 @@ jobs:
       - name: Check style
         run: |
           pipx run pycodestyle examples/*
-          pipx run ruff check --select=ASYNC,RUF006
+          pipx run ruff check
```
cclauss, line 30:
```suggestion
          pipx run ruff check  # default rules
          pipx run ruff check --select=ASYNC,RUF006  # Special rules for asyncio
```
This changes things.  Normally, `ASYNC` and `RUF` are ___off by default___, so this change removes those rules and adds the default rules.  https://docs.astral.sh/ruff/rules  Let's run both just to be safe.

cclauss (changes requested):

.github/workflows/main.yml
julianoes: Oh, I see. Thanks.

examples/camera_params.py
julianoes: Well, so this code was working fine but ruff complained that the tasks were dangling: https://docs.astral.sh/ruff/rules/asyncio-dangling-task/.

So, the advice was to create a strong reference to the tasks. However, now ruff complains that tasks are assigned but never used. It feels like you can't win.

cclauss: I will re-review this evening.

cclauss: I was wrong...  This usage is correct.
```
>>> import asyncio
>>> async def print_char(char: str) -> None:
...     print(char)
...
>>> async def run(chars: str) -> None:
...     tasks = [asyncio.create_task(print_char(char)) for char in chars]
...
>>> asyncio.run(run("abcdefg"))
a
b
c
d
e
f
g
```

examples/camera_params.py
```diff
@@ -24,20 +24,22 @@ async def run():
...
     await drone.connect(system_address="udpin://0.0.0.0:14540")
 
-    asyncio.ensure_future(observe_current_settings(drone))
-    asyncio.ensure_future(observe_camera_mode(drone))
-    asyncio.ensure_future(observe_possible_setting_options(drone))
+    _tasks = [
+        asyncio.create_task(observe_current_settings(drone)),
+        asyncio.create_task(observe_camera_mode(drone)),
+        asyncio.create_task(observe_possible_setting_options(drone)),
+    ]
```
cclauss, line 31: This syntax would delete tasks as they are completed, as discussed in `ruff rule RUF006`.
```suggestion
    tasks = {
        asyncio.create_task(observe_current_settings(drone)),
        asyncio.create_task(observe_camera_mode(drone)),
        asyncio.create_task(observe_possible_setting_options(drone)),
    }
    for task in tasks:
        task.add_done_callback(tasks.discard)
```

julianoes (commented):
I think that's an improvement, let's get it in before I have to fix more conflicts.

PR #787, opened 2025-09-05
Fix typos with codespell and lint Python with ruff

cclauss:
https://github.com/codespell-project/codespell
https://docs.astral.sh/ruff

@julianoes

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -11,6 +11,15 @@ on:
     - '*'
 
 jobs:
+  codespell_and_ruff:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v5
+    - uses: codespell-project/actions-codespell@v2
+      with:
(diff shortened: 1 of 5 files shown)


mavsdk/component_information_pb2.py
```diff
@@ -77,7 +77,7 @@
 _sym_db.RegisterMessage(ComponentInformationResult)
 
 _COMPONENTINFORMATIONSERVICE = DESCRIPTOR.services_by_name['ComponentInformationService']
-if _descriptor._USE_C_DESCRIPTORS == False:
+if _descriptor._USE_C_DESCRIPTORS is False:
```
julianoes, line 80: I don't think we should mess with auto-generated source code.

julianoes (commented):
Thanks @cclauss. I've added a commit to my PR inspired by this.

julianoes:
Incorporated in #786

PR #788, opened 2025-09-06
Remove useless semicolons and trailing whitespace

cclauss:
% `ruff check --select=E703,W291 --fix --unsafe-fixes`
```
Found 26 errors (26 fixed, 0 remaining).
```
% [`ruff rule E703`](https://docs.astral.sh/ruff/rules/useless-semicolon)
% [`ruff rule W291`](https://docs.astral.sh/ruff/rules/trailing-whitespace)


JonasVautherin (approved):

julianoes (changes requested):
> # DO NOT EDIT! This file is auto-generated from ...

julianoes:
This would have to be fixed in the auto-generation.

PR #794, opened 2025-09-10
Ruff rules for performance

cclauss:
Remove the function call overhead of repeatedly calling `list.append()`.

% `ruff check --select=PERF --output-format=concise`
```
mavsdk/camera.py:2188:13: PERF401 Use a list comprehension to create a transformed list
mavsdk/camera.py:2453:13: PERF401 Use a list comprehension to create a transformed list
mavsdk/camera.py:2513:13: PERF401 Use a list comprehension to create a transformed list
mavsdk/component_information.py:363:13: PERF401 Use a list comprehension to create a transformed list
mavsdk/log_files.py:308:13: PERF401 Use a list comprehension to create a transformed list
mavsdk/mission_raw.py:788:13: PERF401 Use a list comprehension to create a transformed list
mavsdk/mission_raw.py:817:13: PERF401 Use a list comprehension to create a transformed list
mavsdk/mission_raw.py:846:13: PERF401 Use a list comprehension to create a transformed list
mavsdk/tune.py:253:13: PERF401 Use a list comprehension to create a transformed list
Found 9 errors.
```
% `ruff rule PERF401`
# manual-list-comprehension (PERF401)

Derived from the **Perflint** linter.

Fix is sometimes available.

## What it does
Checks for `for` loops that can be replaced by a list comprehension.

## Why is this bad?
When creating a transformed list from an existing list using a for-loop,
prefer a list comprehension. List comprehensions are more readable and
more performant.

Using the below as an example, the list comprehension is ~10% faster on
Python 3.11, and ~25% faster on Python 3.10.

Note that, as with all `perflint` rules, this is only intended as a
micro-optimization, and will have a negligible impact on performance in
most cases.

## Example
```python
original = list(range(10000))
filtered = []
for i in original:
    if i % 2:
        filtered.append(i)
```

Use instead:
```python
original = list(range(10000))
filtered = [x for x in original if x % 2]
```

If you're appending to an existing list, use the `extend` method instead:
```python
original = list(range(10000))
filtered.extend(x for x in original if x % 2)
```

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -27,7 +27,7 @@ jobs:
       - name: Check style
         run: |
           pipx run ruff format --check --line-length=100 examples
-          pipx run ruff check --select=ASYNC,RUF006,E,F --line-length=100 examples
+          pipx run ruff check --select=ASYNC,E,F,PERF,RUF006 --line-length=100 examples
           pipx run codespell .
 
       - name: Install prerequisites
--- a/mavsdk/camera.py
+++ b/mavsdk/camera.py
@@ -2183,11 +2183,7 @@ class Camera(AsyncBase):
         if result.result != CameraResult.Result.SUCCESS:
             raise CameraError(result, "list_photos()", component_id, photos_range)
 
-        capture_infos = []
-        for capture_infos_rpc in response.capture_infos:
-            capture_infos.append(CaptureInfo.translate_from_rpc(capture_infos_rpc))
-
-        return capture_infos
+        return [CaptureInfo.translate_from_rpc(capture_infos_rpc) for capture_infos_rpc in response.capture_infos]
 
     async def camera_list(self):
         """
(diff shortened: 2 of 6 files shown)


JonasVautherin:
Again those are generated files, see e.g. [here](https://github.com/mavlink/MAVSDK-Python/blob/36dd005cd6730507d98c0bc2dc4d8b2c96a5d194/mavsdk/camera.py#L1-L3).

```
# -*- coding: utf-8 -*-
# DO NOT EDIT! This file is auto-generated from
# https://github.com/mavlink/MAVSDK-Python/tree/main/other/templates/py
```

JonasVautherin (changes requested):
Again those are generated files, see e.g. [here](https://github.com/mavlink/MAVSDK-Python/blob/36dd005cd6730507d98c0bc2dc4d8b2c96a5d194/mavsdk/camera.py#L1-L3).

```
# -*- coding: utf-8 -*-
# DO NOT EDIT! This file is auto-generated from
# https://github.com/mavlink/MAVSDK-Python/tree/main/other/templates/py
```

JonasVautherin:
> We can postprocess generation with ruff check

Sure, that's not the problem. I'm merely noting that if you want to contribute, you need to contribute to the templates, not to the generated code that says `DO NOT EDIT! This file is auto-generated`.

Respectfully, I can also run a linter. What takes time is to actually contribute the change and then test it. I would be happy to review changes to the templates, given that they are tested. If I need to do most of the work (changing the code and/or testing it), then I'm less enthusiastic 😇.

PR #795, opened 2025-09-10
Remove unnecessary map usage using a list comprehension

cclauss:
We can postprocess generation with ruff check like:
* https://github.com/mavlink/MAVSDK-Python/blob/5717d9e7f196679bfda485b3a2f1df9db18e812c/other/tools/run_protoc.sh#L102

Or we could have done that rewriting in:
 * #792 

---
% `ruff check --select=C417 --fix --unsafe-fixes`
```
Found 24 errors (24 fixed, 0 remaining).
```
% `ruff rule C417`
# unnecessary-map (C417)

Derived from the **flake8-comprehensions** linter.

Fix is sometimes available.

## What it does
Checks for unnecessary `map()` calls with lambda functions.

## Why is this bad?
Using `map(func, iterable)` when `func` is a lambda is slower than
using a generator expression or a comprehension, as the latter approach
avoids the function call overhead, in addition to being more readable.

This rule also applies to `map()` calls within `list()`, `set()`, and
`dict()` calls. For example:

- Instead of `list(map(lambda num: num * 2, nums))`, use
  `[num * 2 for num in nums]`.
- Instead of `set(map(lambda num: num % 2 == 0, nums))`, use
  `{num % 2 == 0 for num in nums}`.
- Instead of `dict(map(lambda v: (v, v ** 2), values))`, use
  `{v: v ** 2 for v in values}`.

## Example
```python
map(lambda x: x + 1, iterable)
```

Use instead:
```python
(x + 1 for x in iterable)
```

## Fix safety
This rule's fix is marked as unsafe, as it may occasionally drop comments
when rewriting the call. In most cases, though, comments will be preserved.

### DIFF
--- a/mavsdk/camera.py
+++ b/mavsdk/camera.py
@@ -286,12 +286,7 @@ class SettingOptions:
             rpcSettingOptions.component_id,
             rpcSettingOptions.setting_id,
             rpcSettingOptions.setting_description,
-            list(
-                map(
-                    lambda elem: Option.translate_from_rpc(elem),
-                    rpcSettingOptions.options,
-                )
-            ),
+            [Option.translate_from_rpc(elem) for elem in rpcSettingOptions.options],
             rpcSettingOptions.is_range,
         )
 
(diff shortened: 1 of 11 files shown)


JonasVautherin (changes requested):
Again again (mentioned e.g. [here](https://github.com/mavlink/MAVSDK-Python/pull/794)), those files are autogenerated.

EDIT: apparently this is done in #796? Should we keep this one open?

PR #796, opened 2025-09-10
Do not generate unnecessary map usage

cclauss:
Fix #795 in code generation instead of postprocessing with `ruff`.
* #795


julianoes:
Can you run and commit the generation in a second commit?

cclauss:
It does not generate anything useful.

julianoes:
@cclauss so then what do you suggest I do with this PR? I'm really confused by all your PRs. Like I understand what you're trying to do but I don't understand why there are several PRs that all seem to do slightly a similar thing.

Yes, I haven't looked into it in detail, but I'm sorry I also don't have infinite time to look into this. I have to pay the bills after all :)

PR #820, opened 2026-03-10
examples: add simulated vision position feed

alireza787b:
## Summary

Adds a simulator-only example that feeds external-vision position data to PX4
through MAVSDK's Mocap plugin.

- reads the simulator's `ground_truth()` global position stream
- converts nearby WGS84 coordinates to a local NED frame
- applies configurable position noise and sample delay
- publishes `VISION_POSITION_ESTIMATE` at 30 Hz
- enables only horizontal and vertical external-position fusion
- monitors PX4 local-position validity without claiming that it proves fusion

The script does not arm, take off, change modes, or otherwise control the
vehicle. It saves and restores the existing `EKF2_EV_CTRL` value.

## Measurement model

MAVSDK's simulator ground-truth stream provides independent position, but not
independent attitude or velocity. The example therefore sends position only.
It does not feed PX4's own EKF estimate back as a measurement, reinterpret NED
velocity as body-frame velocity, or enable yaw fusion without a yaw source.

`VISION_POSITION_ESTIMATE` is used because it is a documented PX4 NED input
that matches the measurements available from this source. A real VIO or motion
capture integration should replace the simulator source and provide every
enabled measurement in its documented frame.

## Delay

`DELAY_MS` delays sample delivery. To evaluate EKF delay compensation, set
`EKF2_EV_DELAY` accordingly and reboot PX4 before running the example.

## Validation

- `ruff format --check --line-length=100 examples`
- `ruff check --select=ASYNC,RUF006,E,F --line-length=100 examples`
- `ruff check --select=PERF --line-length=100 .`
- `codespell .`
- Python bytecode compilation
- WGS84-to-local-NED numerical check
- generated MAVSDK object and protobuf serialization check
- fake ground-truth stream integration check

The branch is rebased on current `main` and contains one commit.

### DIFF
--- a/examples/vision_position_estimate.py
+++ b/examples/vision_position_estimate.py
@@ -1,0 +1,237 @@
+#!/usr/bin/env python3
+
+"""
+Feed simulated external-vision position data to PX4 using MAVSDK.
+
+This simulator-only example reads MAVSDK's ground-truth position stream, converts
+it to a local NED frame, adds configurable noise and delay, and sends it to PX4
+through the Mocap plugin's VISION_POSITION_ESTIMATE API. Only horizontal and
+vertical position are fused because MAVSDK's ground-truth stream does not provide
+independent attitude or velocity.
+
+The script does not arm, take off, or control the vehicle. Fly it separately
+with QGroundControl, RC, a mission, or another MAVSDK client.
+
+Prerequisites:
+    - A PX4 simulator that publishes ground truth
+    - pip install mavsdk
+
+Usage:
+    1. Start PX4 software-in-the-loop simulation.
+    2. Run: python3 vision_position_estimate.py
+    3. Press Ctrl+C to stop.
+
+DELAY_MS delays sample delivery. When testing delay compensation, configure
+EKF2_EV_DELAY accordingly and reboot PX4 before running the example.
+
+For a real system, replace the ground-truth source and local-frame conversion
+with measurements from the VIO or motion-capture system. Enable velocity or yaw
+fusion only when those measurements are supplied in the documented frames.
+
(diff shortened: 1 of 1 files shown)


JonasVautherin (dismissed):
Nice, thanks!

alireza787b:
Thanks for the review Jonas\! Just pushed a small update renaming `quality` → `quality_percent` to match the proto change from your feedback on [#395](https://github.com/mavlink/MAVSDK-Proto/pull/395).

julianoes (dismissed):
Good to be merged? The todos in the PR description aren't ticked, so I'm not so sure.

alireza787b:
Hey Juliano, not yet. You are right that the description there was stale, thanks for flagging it. I just cleaned that up. This example is waiting on the fresh proto PR (`MAVSDK-Proto#404`) and then the clean replacement MAVSDK PR for the mocap wiring. After that I still want to do one final pass on the example itself before calling it merge-ready.

alireza787b:
The upstream mocap dependencies have landed, so I completed the final validation pass and rebased this branch on current `main`.

That pass found two correctness problems in the earlier example: it used PX4's estimated `position_velocity_ned()` as the external measurement source, and it placed NED velocity in body-frame fields. The refreshed example now reads simulator `ground_truth()`, sends only the independent position data available from that source through `VISION_POSITION_ESTIMATE`, and restores the user's original `EKF2_EV_CTRL` value on exit.

This is a substantive rewrite, so the earlier approval should not be treated as covering the new revision. The branch is now one commit, the local style/serialization checks listed in the description pass, and fresh CI is running. A new review would be appreciated when it completes.

PR #826, opened 2026-07-22
Publish as mavsdk-grpc, hand the mavsdk name to the native binding

julianoes:
Prepares this repo for the `mavsdk` PyPI name going to the native binding in the main MAVSDK repo. No runtime changes, and nothing publishes until a tag is pushed.

A tagged build publishes the same sources twice, selected by `MAVSDK_DIST_NAME`:

| Distribution | Imports as | On import |
|---|---|---|
| `mavsdk-grpc` (default) | `mavsdk_grpc` | silent |
| `mavsdk` (legacy) | `mavsdk` | `FutureWarning` |

`mavsdk-grpc` keeps being developed, including for MAVSDK v4. `mavsdk` ships once more carrying a notice.

**Different import names on purpose** — the bit worth sanity-checking. Sharing `mavsdk` would let the two silently overwrite each other whenever both land in one environment (easily reached via transitive deps; pip has no `Conflicts`). Cost: migrating users edit an import, not just a requirements line — acceptable since `pin mavsdk<4` already covers anyone wanting zero changes. No source file changed; the package imports relatively, so `package_dir` remapping was enough.

Two non-obvious details: `FutureWarning` rather than `DeprecationWarning`, which is filtered by default outside `__main__`; and the docs banner overrides sphinx's `document` block, because `page.html` overrides `body` and would win.

Verified locally: both wheels build with the expected `top_level.txt`; imports behave as tabled; docs banner renders on all 45 pages; ruff and codespell pass.

Also adds `twine upload --skip-existing`, since a published version number can never be reused and this doubles the number of uploads that could fail partway.

Note: `TWINE_PASSWORD` must be account-scoped for the release that first creates `mavsdk-grpc` (a project-scoped token cannot create a project). Worth a TestPyPI dry run before tagging.


mavsdk/__init__.py
```diff
@@ -12,6 +13,32 @@
...
+        "now refers to the native MAVSDK binding, which has a different API.\n"
+        "  To carry on with this API, install 'mavsdk-grpc' and use "
+        "'import mavsdk_grpc as mavsdk'. The code is identical, development "
+        "continues there (including releases for MAVSDK v4), and it installs "
+        "under a separate import name so it cannot collide with the native "
+        "binding.\n"
+        "  To freeze instead, pin 'mavsdk<4' (no further updates).\n"
+        "  To migrate, see https://mavsdk.mavlink.io/main/en/python/",
+        FutureWarning,
+        stacklevel=2,
```
julianoes, line 39: @JonasVautherin this is the important warning.

julianoes (commented):
I think it's sane. We'll find out.

